PR #1, opened 2016-03-22
cockroach-go/testserver: testserver using real binaries

mberhault:
After discussion, moving to this repo.
I've also cleaned up node.go a log and collapsed everything into
a single struct (TestServer).

Among other things:
* dump logs on unexpected failures (non-zero exits before Stop())
* cleanup temp dir

Still todo:
* download binary automatically
* use unix socket when (and if) supported by the server

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -1,0 +1,107 @@
+package testserver
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+const (
+	awsBaseURL      = "https://s3.amazonaws.com/cockroach/cockroach"
+	latestSuffix    = "LATEST"
+	localBinaryPath = "/var/tmp"
+)
+
+func binaryName() string {
+	return fmt.Sprintf("cockroach.%s-%s", runtime.GOOS, runtime.GOARCH)
+}
+
+func binaryNameWithSha(sha string) string {
+	return fmt.Sprintf("%s.%s", binaryName(), sha)
+}
+
+func binaryPath(sha string) string {
+	return filepath.Join(localBinaryPath, binaryNameWithSha(sha))
+}
+
+func latestMarkerURL() string {
+	return fmt.Sprintf("%s/%s.%s", awsBaseURL, binaryName(), latestSuffix)
+}
+
+func binaryURL(sha string) string {
+	return fmt.Sprintf("%s/%s.%s", awsBaseURL, binaryName(), sha)
+}
+
+func findLatestSha() (string, error) {
+	markerURL := latestMarkerURL()
+	marker, err := http.Get(markerURL)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


mberhault:
last commit automatically downloads the latest binary for the current OS/ARCH. this is mostly error handling, downloading is hard!

mberhault:
@bdarnell: this is the new naming + aws download.
I've tested it locally in https://github.com/cockroachdb/examples-go/pull/50, and it's quite happy.

testserver/binaries.go
```diff
@@ -0,0 +1,107 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+const (
+	awsBaseURL      = "https://s3.amazonaws.com/cockroach/cockroach"
+	latestSuffix    = "LATEST"
+	localBinaryPath = "/var/tmp"
```
spencerkimball: Do we want to create a subdir for this stuff in `/var/tmp`?

testserver/testserver.go
```diff
@@ -0,0 +1,303 @@
...
+	return ts.stderrBuf.String()
+}
+
+// PGURL parses the node's stdout log for the postgres URL.
+// It returns an actual URL so that callers may modify individual settings.
+// eg:
+//   u := server.PGURL()
+//   u.Path = "mydatabase"
+//   db, err := sql.Open("postgres", u.String())
+// NOTE: it may not be available for a few milliseconds after start,
```
spencerkimball: Let's remove this comment and not return until after we've gotten the port from `stdout`. Callers to this method should just block.

testserver/testserver_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestRunServer(t *testing.T) {
+	ts, err := NewTestServer()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ts.Start()
+	defer ts.Stop()
+
+	time.Sleep(time.Second)
```
spencerkimball: Remove this `Sleep` invocation.

spencerkimball:
LGTM

testserver/testserver.go
mberhault: added WaitForPGURL to loop over stdout and adjusted comment.

testserver/testserver_test.go
mberhault: replaced with call to WaitForPGURL

testserver/binaries.go
mberhault: I'm not sure we need to, though I don't know what the conventions are for `/var/tmp`. Can't really find much.

testserver/testserver.go
spencerkimball: Why add the additional method? Just simplify and have this one which should use a mutex to block and then cache the URL result once it's computed.

mberhault: changed to a goroutine looking for the url instead and stores it permanently.

spencerkimball:
LGTM

PR #2, opened 2016-03-24
Use unix socket in testserver

mberhault:
Remove all the stdout parsing stuff.
PGURL waits for the socket to exist and loops forever until it does.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -23,8 +23,7 @@
 // ts, err := testserver.NewTestServer()
 // err = ts.Start()
 // defer ts.Stop()
-// url := ts.WaitForPGURL()
-// // Check for nil, and make use of URL.
+// url := ts.PGURL()
 //
 // To use, run as follows:
 //   import "github.com/cockroachdb/cockroach-go/testserver"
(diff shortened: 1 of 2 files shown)


testserver/testserver.go
```diff
@@ -108,7 +110,9 @@ func NewTestServer() (*TestServer, error) {
 		cockroachBinary = "cockroach"
 	}
 
-	baseDir, err := ioutil.TempDir("", "cockroach-testserver")
+	// Force "/tmp/" so avoid OSX' really long temp directory names
```
spencerkimball: s/OSX'/OSX's/

testserver/testserver.go
```diff
@@ -118,6 +122,17 @@ func NewTestServer() (*TestServer, error) {
 		return nil, fmt.Errorf("could not create logs directory: %s: %s", logDir, err)
 	}
 
+	options := url.Values{
+		"host": []string{baseDir},
```
spencerkimball: Surprised the pgurl works like this with sockets.

testserver/testserver.go
mberhault: Done

spencerkimball:
LGTM

testserver/testserver.go
mberhault: yeah, it's really gross. That's the reason I've kept the `--socket` flag hidden for now. If there's real demand for it, we can make it more publicly-visible and more in line with postgres, but for now, this is much simpler to use, especially for the dynamic case.

PR #3, opened 2016-03-26
Use 1ns for clock max offset

spencerkimball:
This prevents tests from getting the same time for the first 250ms of
operation, which is pretty confusing. Since this TestServer is a single
node, there's no point in the max offset being set to the default value.


tschottdorf:
Use `1ns` instead. `0ns` disables some code paths.

spencerkimball:
Done

PR #5, opened 2016-03-27
Add Makefile and circle.yaml file

spencerkimball:


tamird:
Should be circle.yml, I think.

mberhault:
LGTM. apparently it accepts both, but yes, `.yml` is what we've been using elsewhere.

spencerkimball:
I fixed the `circle.yml` naming, but it's apparently not finding any tests in here. What gives?

mberhault:
The first circleCI run showed no circle.yml config, but the second one did, it just failed on a "busy" cockroach binary. that may be solved by the issue you just assigned to me

mberhault:
actually, it's not the issue you listed, it's because the tests are being run at the same time, so one of them is trying to run the binary while the other is still downloading it. blast, I need to find a way around that.

spencerkimball:
Should I check this in?

mberhault:
sure. I'll see what I can do to allow concurrent tests when downloading the binary. it's unrelated to this though.

PR #8, opened 2016-03-28
Improve test server

### LINKED ISSUE #6: testserver allows connections before db "system" table is created
This will cause tests to sporadically fail because cockroach receives SQL to create new database / table before it's ready.

### LINKED ISSUE #7: allow concurrent tests with testserver when binary is being downloaded
if two tests using the testserver are running concurrently (often when testing multiple packages), we may end up with one test starting the download, and another thinking the file is there (we only look for filenames).
This obviously causes the second test to fail since the binary file is still busy.

Possible solutions:
* tmp file names for the binary: kindda crappy, we'd have to download it for every single test
* tmp file that gets renamed: the binary would still get downloaded as many times as it takes for the first test to finish.
* lock file until the download is finished: would make all the other tests wait and only download once. any issues if the downloader dies? the test fails anyway.
* initOnce statement somewhere to download: this works if the test runner is a go binary that's running all tests, otherwise not so much.

mberhault:
* use file mode to determine whether it's finished downloading: fixes #7
* loop until system table exists: fixes #6

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/8)

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -10,12 +10,14 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 )
 
 const (
-	awsBaseURL      = "https://s3.amazonaws.com/cockroach/cockroach"
-	latestSuffix    = "LATEST"
-	localBinaryPath = "/var/tmp"
+	awsBaseURL       = "https://s3.amazonaws.com/cockroach/cockroach"
+	latestSuffix     = "LATEST"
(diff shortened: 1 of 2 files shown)


mberhault:
weird failures. after sshing into the machines, the testserver package is actually the old one (even with cache cleared).
These failures will clear once the PR is merged

petermattis:
LGTM

PR #13, opened 2016-04-19
don't overwrite errors if we fail to commit; introduce AmbiguousCommitError.

andreimatei:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/13)

### DIFF
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -24,9 +24,19 @@ import (
 	"github.com/cockroachdb/pq"
 )
 
+// AmbiguousCommitError represents an error that left a transaction in an
+// ambiguous state: unclear if it committed or not.
+type AmbiguousCommitError struct {
+	error
+}
+
 // ExecuteTx runs fn inside a transaction and retries it as needed.
 // On non-retryable failures, the transaction is aborted and rolled
 // back; on success, the transaction is committed.
+// There are cases where the state of a transaction is inherently ambiguous: if
+// we err on RELEASE with a communication error it's unclear if the transaction
+// has been committed or not (similar to erroring on COMMIT in other databases).
+// In that case, we return AmbiguousCommitError.
 //
 // For more information about CockroachDB's transaction model see
 // https://cockroachlabs.com/docs/transactions.html.
@@ -42,9 +52,12 @@ func ExecuteTx(db *sql.DB, fn func(*sql.Tx) error) (err error) {
 	}
 	defer func() {
 		if err == nil {
-			err = tx.Commit()
+			// Ignore commit errors. The tx has already been committed by RELEASE.
+			_ = tx.Commit()
 		} else {
-			tx.Rollback()
+			// We always need to execute a Rollback() so sql.DB releases the
+			// connection.
+			_ = tx.Rollback()
 		}
 	}()
 	// Specify that we intend to retry this txn in case of CockroachDB retryable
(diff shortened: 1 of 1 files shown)


RaduBerinde:
LGTM

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[crdb/tx.go, line 30 \[r1\]](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-KFjokRB_opUdnff6fx5:-KFjokRB_opUdnff6fx6:93740630) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/58b2235e8ade40572890cccea45f75c32e9b646b/crdb/tx.go#L30)):*
if you just do an anonymous `error` you won't need to define the `Error` method

---

*[crdb/tx.go, line 89 \[r1\]](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-KFjpsk7EhdOWsLGnZQi:-KFjpsk7EhdOWsLGnZQj:-646860401) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/58b2235e8ade40572890cccea45f75c32e9b646b/crdb/tx.go#L89)):*
If the `Rollback()` call in the deferred succeeds, would that be a sure indication that the release did not go through? Or do we not expect Rollback to return an error if the release had gone through?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-:-KFjqGSfWbQtPoLzsdSv:676834233)*

crdb/tx.go
```diff
@@ -24,9 +24,26 @@ import (
 	"github.com/cockroachdb/pq"
 )
 
+// AmbiguousCommitError represents an error that has a transaction into an
```
spencerkimball: s/has a transaction into/left a transaction in/

spencerkimball:
LGTM

andreimatei:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[crdb/tx.go, line 27 \[r1\]](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-KFjnfom35zY2ktc_eCj-r1-27:-KFjwSeo7r2e5l60Yc94:-499146009) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/58b2235e8ade40572890cccea45f75c32e9b646b/crdb/tx.go#L27)):*
Done.

---

*[crdb/tx.go, line 30 \[r1\]](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-KFjokRB_opUdnff6fx5:-KFjwGebO01Zi9OgIA19:-499146009) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/58b2235e8ade40572890cccea45f75c32e9b646b/crdb/tx.go#L30)):*
Done.

---

*[crdb/tx.go, line 89 \[r1\]](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-KFjpsk7EhdOWsLGnZQi:-KFjwTPmbOub9-6RKJhL:1381359651) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/58b2235e8ade40572890cccea45f75c32e9b646b/crdb/tx.go#L89)):*
The database accepts a `ROLLBACK` statement even if the `RELEASE` has gone through (analogous to accepting a `COMMIT` after an error - these two statements are always expected to be accepted). It would return a "commit tag" instead of a "rollback tag" so I guess we could theoretically tell if that's the case and indeed conclude that the RELEASE had gone through.
Except we don't have access to that tag. We'd have to use lib/pq directly, and even then I think we don't have access.
But anyway, if `RELEASE` failed with anything but a retryable error, I think there's a fat chance ("fat chance" means "slim chance", right?) that the `tx.Rollback()` will not error (`RELEASE` is not supposed to error). So I don't think there's a point in trying to hard here. I've added a comment to the Rollback() call to explain why we insist on always doing it, regardless of whether we think it'll succeed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13)*

RaduBerinde:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/13#-:-KFk570h6VY_UKA63o2L:1417088721)*

PR #18, opened 2016-09-14
Move the client lib to use the standard lib/pq

### LINKED ISSUE #17: Cannot be used in conjunction with github.com/lib/pq
Attempting to use this for a project where I'm hoping to also support postgres as well as cockroach db. However, since this package imports "github.com/cockroachdb/pq" it's impossible to also register the "github.com/lib/pq" postgres driver. The registration panics because they both want to be the "postgres" driver.

```
panic: sql: Register called twice for driver postgres

goroutine 1 [running]:
panic(0x714d60, 0xc420010d50)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
database/sql.Register(0x76b2ba, 0x8, 0xabda60, 0xaefe88)
	/usr/local/go/src/database/sql/sql.go:45 +0x1b2
github.com/lib/pq.init.1()
	/home/eric/src/github.com/lib/pq/conn.go:44 +0x5c
github.com/lib/pq.init()
	/home/eric/src/github.com/lib/pq/user_posix.go:25 +0x62f
github.com/coreos/dex/storage/sql.init()
```

What's the reason for using the fork? Is there a plan to get any required patches back into the upstream repo?

andreimatei:
... instead of our forked one. This will help people who are trying to
link in both crdb and lib/pq.

The test fix was necessary because our forked libpq has a different way
of consuming results, making exhausting the `rows` cursor somehow not
necessary.

Fixes #17

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/18)

### DIFF
--- a/crdb/tx_test.go
+++ b/crdb/tx_test.go
@@ -50,14 +50,17 @@ INSERT INTO d.t (acct, balance) VALUES (1, 100), (2, 100);
 		if err != nil {
 			return
 		}
-		for i, bal := range []*int{&bal1, &bal2} {
-			if !rows.Next() {
-				err = fmt.Errorf("expected two balances; got %d", i)
+		defer rows.Close()
+		balances := []*int{&bal1, &bal2}
+		i := 0
+		for ; rows.Next(); i += 1 {
+			if err = rows.Scan(balances[i]); err != nil {
 				return
 			}
-			if err = rows.Scan(bal); err != nil {
-				return
-			}
+		}
+		if i != 2 {
+			err = fmt.Errorf("expected two balances; got %d", i)
+			return
 		}
 		return
 	}
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -21,7 +21,7 @@ package crdb
 import (
 	"database/sql"
 
-	"github.com/cockroachdb/pq"
+	"github.com/lib/pq"
 )
 
 // AmbiguousCommitError represents an error that left a transaction in an


crdb/tx_test.go
```diff
@@ -22,7 +22,7 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/cockroachdb/cockroach-go/testserver"
```
tamird: this ain't right

crdb/tx_test.go
```diff
@@ -50,15 +50,18 @@ INSERT INTO d.t (acct, balance) VALUES (1, 100), (2, 100);
 		if err != nil {
 			return
 		}
-		for i, bal := range []*int{&bal1, &bal2} {
-			if !rows.Next() {
-				err = fmt.Errorf("expected two balances; got %d", i)
-				return
-			}
-			if err = rows.Scan(bal); err != nil {
+		defer rows.Close()
```
tamird: i don't understand what you changed here. can you help?

in particular, it seems that if you tried to scan a third value you'd hit an index out of bounds exception.

spencerkimball: What Tamir said.

spencerkimball:
LGTM

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfOpQc2A852G4ngzw4:-KRfOpQc2A852G4ngzw5:binsxlw) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx.go#L24)):*
> ```Go
> 	"database/sql"
> 
> 	"github.com/lib/pq"
> ```

Doesn't this prevent us from using `github.com/cockroachdb/pq` with `cockroach-go`?  I think some of the code in `github.com/cockroachdb/examples-go` will break. And some code in the main cockroach repo which uses `cockroach-go`.

---

*[crdb/tx_test.go, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfPK72NfhCPFg5ds89:-KRfPK72NfhCPFg5ds8A:b2s2ria) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx_test.go#L25)):*
> ```Go
> 	"testing"
> 
> 	"github.com/cockroach-go/testserver"
> ```

What is this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18)*

andreimatei:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfOpQc2A852G4ngzw4:-KRgEc6_NHUJ4_9Ayk89:b-u8q5zw) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx.go#L24)):*
<details><summary><i>Previously, petermattis (Peter Mattis) wrote…</i></summary>
> Doesn't this prevent us from using `github.com/cockroachdb/pq` with `cockroach-go`?  I think some of the code in `github.com/cockroachdb/examples-go` will break. And some code in the main cockroach repo which uses `cockroach-go`.
</details>
yeah, the examples I was fixing.
I didn't realize that the main repo uses crdb too, but indeed it does for some acceptance test. And that sucker transitively depends on `github.com/cockroachdb/pq`. I think I could break that dependency, but I think the better thing to do is to remove all references to `github.com/cockroachdb/pq` from the main repo, except for the one in the CLI. Nobody should need our patches, except the CLI, and nobody depends on the CLI. What do you think?

Btw, I think I once used some tool to give an import chain between two packages, like a recursive `go list`, but now I can't find it. Do y'all happen to have a way?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18)*

andreimatei:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfOpQc2A852G4ngzw4:-KRgGSfQoFkdnJuZ1Wcb:bmf2vvi) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx.go#L24)):*
<details><summary><i>Previously, andreimatei (Andrei Matei) wrote…</i></summary>
> yeah, the examples I was fixing.
> I didn't realize that the main repo uses crdb too, but indeed it does for some acceptance test. And that sucker transitively depends on `github.com/cockroachdb/pq`. I think I could break that dependency, but I think the better thing to do is to remove all references to `github.com/cockroachdb/pq` from the main repo, except for the one in the CLI. Nobody should need our patches, except the CLI, and nobody depends on the CLI. What do you think?
> 
> Btw, I think I once used some tool to give an import chain between two packages, like a recursive `go list`, but now I can't find it. Do y'all happen to have a way?
</details>
This was incomplete. We'd also need to make `github.com/cockroachdb/pq` register the driver under a different name than `postgres`, and have the CLI use that other name for querying.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfOpQc2A852G4ngzw4:-KRhmWQLa1u-WL_6fIH1:bi5qpnt) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx.go#L24)):*
<details><summary><i>Previously, andreimatei (Andrei Matei) wrote…</i></summary>
> This was incomplete. We'd also need to make `github.com/cockroachdb/pq` register the driver under a different name than `postgres`, and have the CLI use that other name for querying.
</details>
I agree with that plan. Only the cli needs to use `cockroachdb/pq`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18)*

andreimatei:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[crdb/tx.go, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfOpQc2A852G4ngzw4:-KRiemR76aHFMgcSXaLu:b-rec3i1) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx.go#L24)):*
<details><summary><i>Previously, petermattis (Peter Mattis) wrote…</i></summary>
> I agree with that plan. Only the cli needs to use `cockroachdb/pq`.
</details>
I went even further - there's no need for our fork to register any name (since the CLI doesn't use it through go's sql package). So I'm making that registration gated on a build tag, and should only be useful for running the library's tests.

So the plan is: land this and the fix to `go-examples` at the same time. Main cockroaches glocks this dependency, so it'll continue to run. Then land the change to main cockroach together with a glock update.

---

*[crdb/tx_test.go, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRfPK72NfhCPFg5ds89:-KRidcm3EaC2DmW6UHxJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx_test.go#L25)):*
<details><summary><i>Previously, petermattis (Peter Mattis) wrote…</i></summary>
> What is this?
</details>
Done.

---

*[crdb/tx_test.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRef0vJ0a6YpPNMqAgv-r1-53:-KRidr-NRnEkZH7vLjyE:b-2jwxsp) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx_test.go#L53)):*
<details><summary><i>Previously, spencerkimball (Spencer Kimball) wrote…</i></summary>
> What Tamir said.
</details>
i changed two things:

1) added the `defer rows.Close()`. This is mandated by the interface, and actually required with the unforked lib/pq, to consume all the results from the socket. Otherwise the connection gets desynchronized. This was sufficient to get the test passing.

2) the change below is just a minor rewrite. I think it's more idiomatic to iterate until `Next()` returns false. It's more defensive too - if the code had been written this way, it would've worked even without the `rows.Close()` call. And also it panics if the query suddenly starts returning 3 values - that's a good thing; think of it as a sanity assertion. Before, we would've silently ignored the 3rd value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18)*

tamird:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[crdb/tx_test.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-KRef0vJ0a6YpPNMqAgv-r1-53:-KRihVaI_wwgFrqbW_Bi:b9ei572) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c355fae646b95bd2a887080a2b6ea5b2a3a7c478/crdb/tx_test.go#L53)):*
<details><summary><i>Previously, andreimatei (Andrei Matei) wrote…</i></summary>
> i changed two things:
> 
> 1) added the `defer rows.Close()`. This is mandated by the interface, and actually required with the unforked lib/pq, to consume all the results from the socket. Otherwise the connection gets desynchronized. This was sufficient to get the test passing.
> 
> 2) the change below is just a minor rewrite. I think it's more idiomatic to iterate until `Next()` returns false. It's more defensive too - if the code had been written this way, it would've worked even without the `rows.Close()` call. And also it panics if the query suddenly starts returning 3 values - that's a good thing; think of it as a sanity assertion. Before, we would've silently ignored the 3rd value.
</details>
I like the old version better. Just add an assertion that `rows.Next()` returns false after the `for .. := range` loop and call it a day.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18)*

petermattis:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/18#-:-KRj0LQZOBEtfWCpS57D:bnfp4nl)*

PR #24, opened 2016-12-20
Expose TCP/IP URL of TestServers

nvanbenschoten:
This is needed in cases where Unix sockets cannot be used, such as with
Java and the PGJDBC driver.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/24)

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -55,10 +55,12 @@
 package testserver
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net/url"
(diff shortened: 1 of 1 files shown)


tamird:
LGTM but let's change the "socketPort" so it's not misleading...and maybe let's rewrite the logging facilities here down the road o.O

bdarnell:
LGTM, but if the TCP port is always going to be bound and available, maybe we should just get rid of the unix socket support.  Unix sockets are nice if they let you avoid all the hackery we've done around TCP sockets, but since we've done it anyway there's little value in keeping them around. 

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[testserver/testserver.go, line 259 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/24#-KZU_06uxpzcIaOK4oTH:-KZU_06viHxC47cRReAP:bjsuzur) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/b71962583b3be9fd6db97f359c05b1ea99748b6e/testserver/testserver.go#L259)):*
> ```Go
> 		time.Sleep(time.Millisecond * 100)
> 	}
> 	return nil, err
> ```

You can move definition of `err` here; we'll never return an error from inside the loop from this path.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/24#-:-KZUZW7pG5QnNWW__Wf_:b-w8syaw)*

nvanbenschoten:
I agree Ben, I just want sure if removing the unix socket support would be possible. I've gone ahead and ripped it out in favor of the standard network URL. PTAL. I'll remove the first commit if we like that approach better.

bdarnell:
Yeah, we can remove the unix socket support - we're not using it anywhere but here (it makes me sad to lose the unix socket option just because java doesn't like posix, but i guess this will also help for our eventual windows port).

Instead of parsing stdout, though, maybe we should add a new option to the server, something like `--listening-url-file` which would specify a filename to which the url would be written once startup is complete. 

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/24#-:-K_ceBftIN6nUHnnpISA:bcwwiyc)*

nvanbenschoten:
Is specifying a filename in a startup flag to receive targetted feedback from a program a common pattern? I don't think I've ever seen it before, but I've also never seen it done any other way either.

bdarnell:
The precedent I had in mind was PID files, which are fairly analogous. There's also the systemd notification protocol, which is similar but uses an environment variable instead of a command-line flag.

nvanbenschoten:
Opened https://github.com/cockroachdb/cockroach/issues/13327 to track this new flag.

tamird:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[testserver/testserver.go, line 358 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/24#-KbqWAc6IFI1FjXGsHhJ:-KbqWAc6IFI1FjXGsHhK:b-yb8p21) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/053408435186c9050ec5b9c78cf07f64c96c8d0f/testserver/testserver.go#L358)):*
> ```Go
> }
> 
> // patternInterceptor wraps a io.Writer and attempts to match the data stream
> ```

did not review below here, fwiw

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/24#-:-KbqWKZPs2dFZ7r3p_tW:bnfp4nl)*

PR #28, opened 2017-05-01
testserver: use content-disposition to determine filename

tamird:
I've verified that net/http doesn't download the entire file until
the response body is consumed (i.e. the call to http.Get is fast, and
the call to io.Copy is slow).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/28)

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"mime"
 	"net/http"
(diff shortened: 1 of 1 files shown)


tamird:
Heh, turns out this code was downloading quite an old binary because I had forgotten to make it track the correct libc-specific key on Linux. We should get this merged ASAP, so that people using this package in the release candidate time frame don't get old binaries.

PR #29, opened 2017-05-03
testserver: look for sql url on stderr

benesch:
The latest version of CockroachDB prints the SQL url on stderr instead of stdout when `--logtostderr` is specified.

@knz, I know you've been working on the logging recently. Let me know if this change wasn't intentional (i.e., if the SQL url should still be printed on stdout when `--logtostderr` is passed).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/29)

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -104,7 +104,7 @@ type TestServer struct {
 	args       []string
 	stdout     string
 	stderr     string
-	stdoutIntr *patternInterceptor
+	stderrIntr *patternInterceptor
 	stdoutBuf  logWriter
 	stderrBuf  logWriter
 }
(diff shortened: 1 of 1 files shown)


knz:
The change here comes from https://github.com/cockroachdb/cockroach/pull/15597

The complaint (from Alex, Andrei) was that the node startup message with this info was printed both on stdout and stderr when --logtostderr is enabled.

The reasoning behind was that 1) the startup message must be copied to the logs always (for later investigation) and 2) for consistency it would be strange to have different logging output depending on whether --logtostderr is set 

So I decided to not place it to stdout any more, but now your issue highlights that this may have been the wrong call -- perhaps stdout is the right place after all? 
Suggestions welcome.

benesch:
I found a better way; submitted as #30.

PR #31, opened 2017-05-03
testserver: use glibc binary on systems with eglibc

benesch:
Some versions of Ubuntu use eglibc ("embedded glibc") instead of glibc,
so they don't match the glibc regex.  Attempting to run the musl
binaries on these systems hangs forever. (We should probably look into
this separately; the musl binaries should work on these systems.) For
now, though, work around the issue by grepping for the less-common
"musl" rather than "glibc."

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/31)

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -38,7 +38,7 @@ func downloadFile(response *http.Response, filePath string) error {
 	return os.Chmod(filePath, finishedFileMode)
 }
 
-var glibcRE = regexp.MustCompile(`(?i)\bglibc\b`)
+var muslRE = regexp.MustCompile(`(?i)\bmusl\b`)
 
 func downloadLatestBinary() (string, error) {
 	goos := runtime.GOOS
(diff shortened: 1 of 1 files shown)


dt:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/31#-:-KjEhO3Ho20vxx_qhocb:bnfp4nl)*

tamird:
From slack, for posterity:

Thanks for finding and fixing this! The reason I was looking for gnu is that musl is fully static and should run anywhere. Bummer that it hangs, good thing we're not publishing those binaries anywhere. We should flip the logic when the musl build actually works though. I think this is https://github.com/jemalloc/jemalloc/issues/585.

Filed as #32.

PR #34, opened 2017-05-08
testserver: ensure no whitespace in URLs

tamird:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/34)

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -55,6 +55,7 @@
 package testserver
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
 	"flag"
@@ -263,7 +264,7 @@ func (ts *TestServer) pollListeningURLFile() error {
 		time.Sleep(100 * time.Millisecond)
 	}
 
-	u, err := url.Parse(string(data))
+	u, err := url.Parse(string(bytes.TrimSpace(data)))
 	if err != nil {
 		return fmt.Errorf("failed to parse SQL URL: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


testserver/testserver_test.go
```diff
@@ -26,8 +27,21 @@ func TestRunServer(t *testing.T) {
...
+
+func TestPGURLWhitespace(t *testing.T) {
+	ts, err := testserver.NewTestServer()
 	if err != nil {
 		t.Fatal(err)
 	}
+	if err := ts.Start(); err != nil {
+		t.Fatal(err)
+	}
+	url := ts.PGURL().String()
```
jordanlewis: when would this return a url with a space in the first place?

testserver/testserver.go
```diff
@@ -263,7 +264,7 @@ func (ts *TestServer) pollListeningURLFile() error {
 		time.Sleep(100 * time.Millisecond)
 	}
 
-	u, err := url.Parse(string(data))
+	u, err := url.Parse(string(bytes.TrimSpace(data)))
```
jordanlewis: should we urlencode the spaces?

testserver/testserver_test.go
tamird: Yep! The test fails before this change.

testserver/testserver.go
tamird: In this case we're just seeing the trailing newline in the url file. If you can help me conceive of a test that would exercise legitimate spaces in the URL I'd be happy to adding urlencoding.

tamird:
Discussed in person, merging.

PR #35, opened 2017-06-11
Add ExecuteInTx

adamgee:
Add support for executing cockroach transactions using [GORM](https://github.com/jinzhu/gorm).

Refactor retry logic into internal function which uses an internal 
`transaction` interface.

### DIFF
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -60,8 +60,10 @@ type Tx interface {
 }
 
 // ExecuteInTx runs fn inside tx which should already have begun.
-// *** WARNING ***: tx should not have been used before being passed to ExecuteInTx; any
-// statements executed prior may be rolled back even though tx may commit successfully.
+// WARNING: Do not execute any statements on the supplied tx before calling this function.
+// ExecuteInTx will only retry statements that are performed within the supplied
+// closure (fn). Any statements performed on the tx before ExecuteInTx is invoked will *not*
+// be re-run if the transaction needs to be retried.
 func ExecuteInTx(tx Tx, fn func() error) (err error) {
 	defer func() {
 		if err == nil {
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -60,7 +60,7 @@ type Tx interface {
 }
 
 // ExecuteInTx runs fn inside tx which should already have begun.
-// WARNING: Do not execute any statements on the supplied tx before calling this function.
+// *WARNING*: Do not execute any statements on the supplied tx before calling this function.
 // ExecuteInTx will only retry statements that are performed within the supplied
 // closure (fn). Any statements performed on the tx before ExecuteInTx is invoked will *not*
 // be re-run if the transaction needs to be retried.
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -21,6 +21,7 @@ package crdb
 import (
 	"database/sql"
 
+	"github.com/jinzhu/gorm"
 	"github.com/lib/pq"
 )
 
@@ -50,6 +51,40 @@ func ExecuteTx(db *sql.DB, fn func(*sql.Tx) error) (err error) {
 	if err != nil {
 		return err
 	}
+	return executeTxInternal(tx, func() error { return fn(tx) })
+}
+
+func ExecuteGORMTx(db *gorm.DB, fn func(*gorm.DB) error) (err error) {
+	tx := db.Begin()
+	if tx.Error != nil {
+		return tx.Error
+	}
+	return executeTxInternal(&gormTx{tx}, func() error { return fn(tx) })
+}
+
+type transaction interface {
+	Exec(query string, args ...interface{}) (sql.Result, error)
+	Commit() error
+	Rollback() error
+}
+
+type gormTx struct {
+	gormDB *gorm.DB
+}
+
+func (db *gormTx) Exec(query string, args ...interface{}) (sql.Result, error) {
+	return nil, db.gormDB.Exec(query, args).Error
+}
+
+func (db *gormTx) Commit() error {
+	return db.gormDB.Commit().Error
+}
+
+func (db *gormTx) Rollback() error {
+	return db.gormDB.Rollback().Error
+}
+
+func executeTxInternal(tx transaction, fn func() error) (err error) {
 	defer func() {
 		if err == nil {
 			// Ignore commit errors. The tx has already been committed by RELEASE.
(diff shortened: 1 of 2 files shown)


crdb/tx.go
```diff
@@ -21,6 +21,7 @@ package crdb
 import (
 	"database/sql"
 
+	"github.com/jinzhu/gorm"
```
petermattis: The addition of the dependency on gorm here is unfortunate and a bit undesirable. I wonder if there is a way to support gorm without adding such a dependency.

adamgee: Wasn't sure how "official" `gorm` was and what the appetite for something like this would be, happy to refactor if this is on the line.  Would the introduction of more generic public `interface`s be preferable: one "Database" with which we could `Begin` a transaction and then the existing `Transaction` interface to `Execute`, `Commit`, and `Rollback`?

petermattis: `gorm` is relatively new, but seems to be the current hotness. I would be hesitant to lock in support for it. A more generic interface would be appreciated. I think that could be achieved by making `executeTxInternal` a public interface and then require users of ORMs such as `gorm` to provide an implementation of `transaction` (as you've done in this PR).

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[crdb/tx.go, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/35#-KmNwMYAtC8fJE1hmGj2-r1-24:-KmSUkt0Hz8UKALszcP8:b-hsic2x) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/0c89d8b3b7886c77823254d8d12c4026f77fd7b9/crdb/tx.go#L24)):*
<details><summary><i>Previously, petermattis (Peter Mattis) wrote…</i></summary><blockquote>

`gorm` is relatively new, but seems to be the current hotness. I would be hesitant to lock in support for it. A more generic interface would be appreciated. I think that could be achieved by making `executeTxInternal` a public interface and then require users of ORMs such as `gorm` to provide an implementation of `transaction` (as you've done in this PR).
</blockquote></details>

We shouldn't depend on `gorm` here, but it might be OK to do so in a different subdirectory of this repo. I think the tools generally track dependencies at the package/directory level instead of the repo level (@dt is that true of `dep`?).

It's unfortunate that all of `gorm.DB`'s methods return a `*gorm.DB`. If not for that, you could mock this out with interfaces without introducing a compile-time dependency on gorm. I wonder if the gorm maintainers could be convinced to add methods like the three you define below so that we wouldn't need to name any gorm types here. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/35)*

dt:
As I understand it, if all the imports of `gorm` are confined to a (sub)package, only those who actually use that package will be forced to pickup the transitive dependency, so we probably should confine these to their own package.

If we vendor gorm (we probably should, so that this can build on its own?), we should also document that it has to be used with a vendor-flattening tool, like `dep` or `glide -v`, to avoid the mismatched symbols.

adamgee:
PTAL

crdb/tx.go
```diff
@@ -50,6 +50,17 @@ func ExecuteTx(db *sql.DB, fn func(*sql.Tx) error) (err error) {
...
+}
+
+type Tx interface {
+	Exec(query string, args ...interface{}) (sql.Result, error)
+	Commit() error
+	Rollback() error
+}
+
+// ExecuteBegunTx runs fn inside tx which should already have begun.
+func ExecuteBegunTx(tx Tx, fn func() error) (err error) {
```
petermattis: I'd like to find a better name than this. `Begun` is an odd term. Perhaps `ExecuteWithTx` or `ExecuteInTx`.

adamgee:
@petermattis PTAL

crdb/tx.go
```diff
@@ -50,6 +50,17 @@ func ExecuteTx(db *sql.DB, fn func(*sql.Tx) error) (err error) {
...
+	return ExecuteInTx(tx, func() error { return fn(tx) })
+}
+
+type Tx interface {
+	Exec(query string, args ...interface{}) (sql.Result, error)
+	Commit() error
+	Rollback() error
+}
+
+// ExecuteInTx runs fn inside tx which should already have begun.
```
petermattis: What happens if statements have already been executed in the transaction before `ExecuteInTx` is called? I don't think those statements will be re-executed. Not sure who on the Cockroach side would know. @bdarnell?

bdarnell: I don't know anything about gorm, but I don't see how those statements could get re-executed. It must be a requirement that the transaction have not been used (except maybe for one-time-only operations like setting the isolation level) before being passed into this function.

crdb/tx_test.go
```diff
@@ -25,9 +25,25 @@ import (
 	"github.com/cockroachdb/cockroach-go/testserver"
 )
 
+func TestExecuteTx(t *testing.T) {
+	executeTx := func(db *sql.DB, fn func(tx *sql.Tx) error) (err error) { return ExecuteTx(db, fn) }
+	testExecuteTxInner(t, executeTx)
+}
+
+func TestExecuteBegunTx(t *testing.T) {
```
bdarnell: Rename the test too.

crdb/tx.go
adamgee: I looked at the sql.Tx and didn't see any way to test for statements already being executed.  I can't think of any way to enforce this, I'm afraid the comment is going to have to do...

Any suggestions on how to phrase this?

adamgee:
@bdarnell @petermattis - PTAL

crdb/tx.go
petermattis:
```
// WARNING: Do not execute any statements on the supplied tx before calling this function.
// ExecuteInTx will only retry statements that are performed within the supplied
// closure (fn). Any statements performed on the tx before ExecuteInTx is invoked will *not* 
// be re-run if the transaction needs to be retried. 
```

bdarnell:
LGTM with Peter's suggestion for the warning text.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/35#-:-KmXciuFSTmhVXomh1hG:b2oox6e)*

adamgee:
@bdarnell  PTAL

adamgee:
@petermattis @bdarnell - PTAL

PR #41, opened 2017-08-07
Add support for context

garvitjuniwal:
Clients may want to use sql package context
functionality, for example, to trace all sql
queries.

### DIFF
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -17,6 +17,7 @@
 package crdb
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 
@@ -42,19 +43,18 @@ type AmbiguousCommitError struct {
 //
 // NOTE: the supplied exec closure should not have external side
 // effects beyond changes to the database.
-func ExecuteTx(db *sql.DB, fn func(*sql.Tx) error) (err error) {
+func ExecuteTx(ctx context.Context, db *sql.DB, txopts *sql.TxOptions, fn func(*sql.Tx) error) error {
 	// Start a transaction.
-	var tx *sql.Tx
-	tx, err = db.Begin()
+	tx, err := db.BeginTx(ctx, txopts)
 	if err != nil {
 		return err
 	}
-	return ExecuteInTx(tx, func() error { return fn(tx) })
+	return ExecuteInTx(ctx, tx, func() error { return fn(tx) })
 }
 
 // Tx is used to permit clients to implement custom transaction logic.
 type Tx interface {
-	Exec(query string, args ...interface{}) (sql.Result, error)
+	ExecContext(context.Context, string, ...interface{}) (sql.Result, error)
 	Commit() error
 	Rollback() error
 }
@@ -64,7 +64,7 @@ type Tx interface {
 // ExecuteInTx will only retry statements that are performed within the supplied
 // closure (fn). Any statements performed on the tx before ExecuteInTx is invoked will *not*
 // be re-run if the transaction needs to be retried.
-func ExecuteInTx(tx Tx, fn func() error) (err error) {
+func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 	defer func() {
 		if err == nil {
 			// Ignore commit errors. The tx has already been committed by RELEASE.
@@ -77,7 +77,7 @@ func ExecuteInTx(tx Tx, fn func() error) (err error) {
 	}()
 	// Specify that we intend to retry this txn in case of CockroachDB retryable
 	// errors.
-	if _, err = tx.Exec("SAVEPOINT cockroach_restart"); err != nil {
+	if _, err = tx.ExecContext(ctx, "SAVEPOINT cockroach_restart"); err != nil {
 		return err
 	}
 
@@ -88,7 +88,7 @@ func ExecuteInTx(tx Tx, fn func() error) (err error) {
 			// RELEASE acts like COMMIT in CockroachDB. We use it since it gives us an
 			// opportunity to react to retryable errors, whereas tx.Commit() doesn't.
 			released = true
-			if _, err = tx.Exec("RELEASE SAVEPOINT cockroach_restart"); err == nil {
+			if _, err = tx.ExecContext(ctx, "RELEASE SAVEPOINT cockroach_restart"); err == nil {
 				return nil
 			}
 		}
@@ -103,7 +103,7 @@ func ExecuteInTx(tx Tx, fn func() error) (err error) {
 			}
 			return err
 		}
-		if _, err = tx.Exec("ROLLBACK TO SAVEPOINT cockroach_restart"); err != nil {
+		if _, err = tx.ExecContext(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); err != nil {
 			// ROLLBACK TO SAVEPOINT failed. If it failed with a lib/pq error, we want
 			// to pass this error to the client, but also include the original error
 			// message and code. So, we'll do some surgery on lib/pq errors in
--- a/crdb/tx_test.go
+++ b/crdb/tx_test.go
@@ -15,6 +15,7 @@
 package crdb
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"sync"
@@ -28,19 +29,32 @@ func TestExecuteTx(t *testing.T) {
 }
 
 func TestExecuteInTx(t *testing.T) {
-	executeTx := func(db *sql.DB, fn func(tx *sql.Tx) error) (err error) {
-		tx, err := db.Begin()
+	executeTx := func(
+		ctx context.Context,
+		db *sql.DB,
+		opts *sql.TxOptions,
+		fn func(*sql.Tx) error,
+	) error {
+		tx, err := db.BeginTx(ctx, opts)
 		if err != nil {
 			return err
 		}
-		return ExecuteInTx(tx, func() error { return fn(tx) })
+		return ExecuteInTx(ctx, tx, func() error { return fn(tx) })
 	}
 	testExecuteTxInner(t, executeTx)
 }
 
 // TestExecuteTx verifies transaction retry using the classic
 // example of write skew in bank account balance transfers.
-func testExecuteTxInner(t *testing.T, executeTxFn func(db *sql.DB, fn func(tx *sql.Tx) error) error) {
+func testExecuteTxInner(
+	t *testing.T,
+	executeTxFn func(
+		context.Context,
+		*sql.DB,
+		*sql.TxOptions,
+		func(*sql.Tx) error,
+	) error,
+) {
 	db, stop := testserver.NewDBForTest(t)
 	defer stop()
 
@@ -82,7 +96,7 @@ INSERT INTO d.t (acct, balance) VALUES (1, 100), (2, 100);
 		errCh := make(chan error, 1)
 		go func() {
 			*iter = 0
-			errCh <- executeTxFn(db, func(tx *sql.Tx) error {
+			errCh <- executeTxFn(context.Background(), db, nil, func(tx *sql.Tx) error {
 				*iter++
 				bal1, bal2, err := getBalances(tx)
 				if err != nil {


tamird:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 52 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-KqxcQeMnZpfkR7eVuvo:-KqxcLYpba9EvjLEaMYD:be7e8cg) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1e844438fc614489730736a95f9f303e8f6623e8/crdb/tx.go#L52)):*
> ```Go
> }
> 
> func ExecuteTxContext(ctx context.Context, db *sql.DB, fn func(*sql.Tx) error) (err error) {
> ```

Perhaps this should take a `*sql.TxOptions` to pass through to `(*sql.DB).BeginTx`.

Also, not your code but while you're here, would you get rid of this named return?

---

*[crdb/tx.go, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-KqxdH2T8I-ZYV3oEM9c:-KqxdH2T8I-ZYV3oEM9d:b-a8ddmw) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1e844438fc614489730736a95f9f303e8f6623e8/crdb/tx.go#L68)):*
> ```Go
> }
> 
> type TxContext interface {
> ```

why do we need this interface? also, why do we have the `Tx` interface? and why do we need `ignoreContext`?

-- looks like these interfaces exist to support older versions of Go by not depending on concrete methods on `*sql.DB`. We should not do this, particularly in the absence of tests, since we get no type safety. Perhaps this should be done with build tags, as is common practice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

garvitjuniwal:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[crdb/tx.go, line 52 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-KqxcQeMnZpfkR7eVuvo:-Kr-goGWfaglFRA7Y0QE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1e844438fc614489730736a95f9f303e8f6623e8/crdb/tx.go#L52)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

Perhaps this should take a `*sql.TxOptions` to pass through to `(*sql.DB).BeginTx`.

Also, not your code but while you're here, would you get rid of this named return?
</blockquote></details>

Done.

---

*[crdb/tx.go, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-KqxdH2T8I-ZYV3oEM9c:-Kr-fzD7eW6olKS707OI:bykvhot) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1e844438fc614489730736a95f9f303e8f6623e8/crdb/tx.go#L68)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

why do we need this interface? also, why do we have the `Tx` interface? and why do we need `ignoreContext`?

-- looks like these interfaces exist to support older versions of Go by not depending on concrete methods on `*sql.DB`. We should not do this, particularly in the absence of tests, since we get no type safety. Perhaps this should be done with build tags, as is common practice.
</blockquote></details>

I added comments in the code.

The Tx interface was added by #35 specifically because adam wanted to use this logic with Gorm txns (which is a wrapper over sql.DB)

I added a separate TxContext because I was trying to be backward compatible with clients that might be depending on the Tx interface. If this is not a concern, I can just merge Tx and TxContext. We are one of the clients but I can fix the stuff on our end. Not sure if there are more clients.

ignoreContext exists so that I can reuse the same code for someone who only implements Tx by just calling Exec from ExecContext (and ignoring the ctx)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

tamird:
LGTM - could you add a simple test so that we can confirm that `sql.Tx` implements `TxContext`?

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[crdb/tx.go, line 52 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-KqxcQeMnZpfkR7eVuvo:-Kr1TrpycyAfauy9XI3p:b-9z7i1i) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1e844438fc614489730736a95f9f303e8f6623e8/crdb/tx.go#L52)):*
> Also, not your code but while you're here, would you get rid of this named return?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-:-Kr1U7c3jDlD-Zb7--9X:b-ng1xc6)*

andreimatei:
My vote is to break backwards compatibility and just change the API to get the context. Seems a bit too early in the life of this library to worry about compatibility - and people who care have probably vendored it.

Also a question - what is the "trace all sql queries" functionality that you speak of in the commit msg? Is that a thing?

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-:-Kr1VhHFxH7GzVj7sp7N:bwjknn0)*

garvitjuniwal:
Broke backwards compatibility and added context everywhere.

Re: tracing, I could use https://github.com/ExpansiveWorlds/instrumentedsql along with this change to produce graphs like the one attached in zipkin.


---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[crdb/tx.go, line 52 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-KqxcQeMnZpfkR7eVuvo:-Kr28bkqfq-4bzaLIscG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1e844438fc614489730736a95f9f303e8f6623e8/crdb/tx.go#L52)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

> Also, not your code but while you're here, would you get rid of this named return?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-:-Kr28eZbnAr9JH6vZhQK:b-aiw62v)*

garvitjuniwal:
Re: making sure sql.Tx implements Tx. This is being checked in the functions ExecuteTx and ExecuteTxContext

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-:-Kr29_DouGlMRK3GKqq3:b-59hod3)*

tamird:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 48 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2AR5JkISZ9qNnm3J1:-Kr2AR5JkISZ9qNnm3J2:b-emttcp) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L48)):*
> ```Go
> // NOTE: the supplied exec closure should not have external side
> // effects beyond changes to the database.
> func ExecuteTx(db *sql.DB, fn func(*sql.Tx) error) error {
> ```

Let's remove this method, since we've already broken backwards-compatibility.

---

*[crdb/tx.go, line 55 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr29nfYiDAvGIUX9uxc:-Kr29nfYiDAvGIUX9uxd:b-r2vraz) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L55)):*
> txopts *sql.TxOptions

FWIW, the presence of this parameter already breaks compatibility since the type doesn't exist before Go 1.8.

I think we should preserve Go 1.7 support for now, but as I said earlier, it's going to require build tags, so I won't insist on it.

---

*[crdb/tx.go, line 67 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2AI9woZVM8VaALbnW:-Kr2AI9woZVM8VaALbnX:brc1bhs) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L67)):*
> ```Go
> // own transaction engine (other than what is provided by package sql.)
> type Tx interface {
> 	Exec(query string, args ...interface{}) (sql.Result, error)
> ```

this can be removed

---

*[crdb/tx.go, line 68 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2AJZc5o6rGus-SxWg:-Kr2AJZc5o6rGus-SxWh:buackmz) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L68)):*
> ```Go
> type Tx interface {
> 	Exec(query string, args ...interface{}) (sql.Result, error)
> 	ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error)
> ```

naming all the arguments is not necessary (you can just write `ExecContext(context.Context, string, ...interface{}) (sql.Result, error)`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

garvitjuniwal:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[crdb/tx.go, line 48 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2AR5JkISZ9qNnm3J1:-Kr2CC4ZseM6Zt6u9PEM:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L48)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

Let's remove this method, since we've already broken backwards-compatibility.
</blockquote></details>

Done.

---

*[crdb/tx.go, line 55 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr29nfYiDAvGIUX9uxc:-Kr2CPe_h4YFfT14uKun:bp2xuk7) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L55)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

> txopts *sql.TxOptions

FWIW, the presence of this parameter already breaks compatibility since the type doesn't exist before Go 1.8.

I think we should preserve Go 1.7 support for now, but as I said earlier, it's going to require build tags, so I won't insist on it.
</blockquote></details>

is it ok if I leave this to be fixed in a follow up commit if necessary?

---

*[crdb/tx.go, line 67 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2AI9woZVM8VaALbnW:-Kr2EXi54Fxlu5_SN4Og:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L67)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

this can be removed
</blockquote></details>

Done.

---

*[crdb/tx.go, line 68 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2AJZc5o6rGus-SxWg:-Kr2EVzow-FUepoVafXL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L68)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

naming all the arguments is not necessary (you can just write `ExecContext(context.Context, string, ...interface{}) (sql.Result, error)`)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

tamird:
Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[crdb/tx.go, line 55 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr29nfYiDAvGIUX9uxc:-Kr2FwnYZYebHS_1dUlL:bb77ucv) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/188e7613662e666670e52827bc9a1374ab921ee8/crdb/tx.go#L55)):*
<details><summary><i>Previously, garvitjuniwal (Garvit Juniwal) wrote…</i></summary><blockquote>

is it ok if I leave this to be fixed in a follow up commit if necessary?
</blockquote></details>

Yep.

---

*[crdb/tx_test.go, line 38 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2F5UrKcLuhSL-CV47:-Kr2F5Usw6JbX2Prti0U:br88n7d) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/43d0f1cdbb3caeb3769eb569ee8a9432ce6c6a06/crdb/tx_test.go#L38)):*
> ```Go
> 		db *sql.DB,
> 		opts *sql.TxOptions,
> 		fn func(tx *sql.Tx) error,
> ```

since you're in here shuffling things around, want to unname this parameter too? `fn func(*sql.Tx) error` - ditto below

---

*[crdb/tx_test.go, line 57 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2FuJmOkvIQ57lGiqt:-Kr2FuJmOkvIQ57lGiqu:bps1s89) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/43d0f1cdbb3caeb3769eb569ee8a9432ce6c6a06/crdb/tx_test.go#L57)):*
> ```Go
> 		db *sql.DB,
> 		opts *sql.TxOptions,
> 		fn func(tx *sql.Tx) error,
> ```

here too

---

*[crdb/tx_test.go, line 101 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2FQVunf2uOmXfA1pb:-Kr2FQVunf2uOmXfA1pc:bps1s89) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/43d0f1cdbb3caeb3769eb569ee8a9432ce6c6a06/crdb/tx_test.go#L101)):*
> ```Go
> 		go func() {
> 			*iter = 0
> 			errCh <- executeTxFn(context.Background(), db, nil, func(tx *sql.Tx) error {
> ```

here too

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

garvitjuniwal:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[crdb/tx_test.go, line 38 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2F5UrKcLuhSL-CV47:-Kr2HSHYVp-z7ZJd_oYP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/43d0f1cdbb3caeb3769eb569ee8a9432ce6c6a06/crdb/tx_test.go#L38)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

since you're in here shuffling things around, want to unname this parameter too? `fn func(*sql.Tx) error` - ditto below
</blockquote></details>

Done.

---

*[crdb/tx_test.go, line 57 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2FuJmOkvIQ57lGiqt:-Kr2HTmZlJ7OP7MVPYi1:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/43d0f1cdbb3caeb3769eb569ee8a9432ce6c6a06/crdb/tx_test.go#L57)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

here too
</blockquote></details>

Done.

---

*[crdb/tx_test.go, line 101 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41#-Kr2FQVunf2uOmXfA1pb:-Kr2HUKKyvDimJ1swryL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/43d0f1cdbb3caeb3769eb569ee8a9432ce6c6a06/crdb/tx_test.go#L101)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

here too
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

tamird:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

tamird:
LGTM but oh dear, tests failed because our CI uses Go 1.7.

@jordanlewis should we just put this on travis? idk why we use TC here.

jordanlewis:
I've never used Travis, so I can't say what the tradeoffs might be. I don't have any objections if you'd like to shepherd that process though.

tamird:
@garvitjuniwal could you rebase over #42?

tamird:
Reviewed 2 of 2 files at r6.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/41)*

tamird:
Thanks!

Issue #16, opened 2016-07-20
should use `errors.Cause(err)` instead of `err` to check for retryable error


tschottdorf:
to allow this code to work correctly:

``` go
if err := tx.Exec("...."); err != nil {
    return errors.Wrap(err, "during insert")
}
```

andreimatei:
I sympathize with the desire, but I don't think we should add a dependency on the `pkg/errors` package.
What I'd do is comment on the `ExecuteTx` to tell people that they can't do that. The need for a comment explaining what errors the closure should return has already arisen.

PR #43, opened 2017-10-09
Wrap non-pq error on rollback

neeral:
If an error occurs attempting to rollback and that error isn't a libpq
error, then the error is currently overwritten by a string. This causes
it to lose its structure and any context it had. Instead, wrapping the
error allows callers to be able to deconstruct the error.

### DIFF
--- a/crdb/error.go
+++ b/crdb/error.go
@@ -1,0 +1,51 @@
+package crdb
+
+import "fmt"
+
+type txError struct {
+	cause error
+}
+
+// Error implements the error interface
+func (e *txError) Error() string { return e.cause.Error() }
+
+// Cause returns the error encountered by the "ROLLBACK TO SAVEPOINT
+// cockroach_restart" statement. This method also implements the internal
+// pkg/errors.causer interface, so TxnRestartError works with pkg/errors.Cause().
+func (e *txError) Cause() error { return e.cause }
+
+// AmbiguousCommitError represents an error that left a transaction in an
+// ambiguous state: unclear if it committed or not.
+type AmbiguousCommitError struct {
+	txError
+}
+
+func newAmbiguousCommitError(err error) *AmbiguousCommitError {
+	return &AmbiguousCommitError{txError{cause: err}}
+}
+
+// TxnRestartError represents an error when restarting a transaction. `cause` is
+// the error from restarting the txn and `retryCause` is the original error which
+// triggered the restart.
+type TxnRestartError struct {
+	txError
+	retryCause error
+	msg        string
+}
+
+func newTxnRestartError(err error, retryErr error) *TxnRestartError {
+	const msgPattern = "restarting txn failed. ROLLBACK TO SAVEPOINT " +
+		"encountered error: %s. Original error: %s."
+	return &TxnRestartError{
+		txError:    txError{cause: err},
+		retryCause: retryErr,
+		msg:        fmt.Sprintf(msgPattern, err, retryErr),
+	}
+}
+
+// Error implements the error interface
+func (e *TxnRestartError) Error() string { return e.msg }
+
+// RetryCause returns the error encountered by the transaction, which caused the
+// transaction to be restarted.
+func (e *TxnRestartError) RetryCause() error { return e.retryCause }
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -19,16 +19,9 @@ package crdb
 import (
 	"context"
 	"database/sql"
-	"fmt"
 
 	"github.com/lib/pq"
 )
-
-// AmbiguousCommitError represents an error that left a transaction in an
-// ambiguous state: unclear if it committed or not.
-type AmbiguousCommitError struct {
-	error
-}
 
 // ExecuteTx runs fn inside a transaction and retries it as needed.
 // On non-retryable failures, the transaction is aborted and rolled
@@ -37,11 +30,13 @@ type AmbiguousCommitError struct {
 // we err on RELEASE with a communication error it's unclear if the transaction
 // has been committed or not (similar to erroring on COMMIT in other databases).
 // In that case, we return AmbiguousCommitError.
+// There are cases when restarting a transaction fails: we err on ROLLBACK
+// to the SAVEPOINT. In that case, we return a TxnRestartError.
 //
 // For more information about CockroachDB's transaction model see
 // https://cockroachlabs.com/docs/stable/transactions.html.
 //
-// NOTE: the supplied exec closure should not have external side
+// NOTE: the supplied fn closure should not have external side
 // effects beyond changes to the database.
 func ExecuteTx(ctx context.Context, db *sql.DB, txopts *sql.TxOptions, fn func(*sql.Tx) error) error {
 	// Start a transaction.
@@ -99,24 +94,12 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 		pqErr, ok := err.(*pq.Error)
 		if retryable := ok && (pqErr.Code == "CR000" || pqErr.Code == "40001"); !retryable {
 			if released {
-				err = &AmbiguousCommitError{err}
+				err = newAmbiguousCommitError(err)
 			}
 			return err
 		}
-		if _, err = tx.ExecContext(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); err != nil {
-			// ROLLBACK TO SAVEPOINT failed. If it failed with a lib/pq error, we want
-			// to pass this error to the client, but also include the original error
-			// message and code. So, we'll do some surgery on lib/pq errors in
-			// particular.
-			// If it failed with any other error (e.g. the "driver: bad connection" is
-			// untyped), we overwrite the error.
-			msgPattern := "restarting txn failed. ROLLBACK TO SAVEPOINT encountered error: %s. " +
-				"Original error (code: %s): %s."
-			if rollbackPQErr, ok := err.(*pq.Error); ok {
-				rollbackPQErr.Message = fmt.Sprintf(msgPattern, rollbackPQErr, pqErr.Code, pqErr)
-				return rollbackPQErr
-			}
-			return fmt.Errorf(msgPattern, err, pqErr.Code, pqErr)
+		if _, retryErr := tx.ExecContext(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); retryErr != nil {
+			return newTxnRestartError(retryErr, err)
 		}
 	}
 }


vivekmenezes:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw2qiaaCEZKq8YBPtRr:bnfp4nl)*

andreimatei:
I don't think we should add a dependency on pkg/errors to this library.
This has been discussed before in https://github.com/cockroachdb/cockroach/issues/17667 and in #16

For the particular thing you're changing, what prompted you to do this? What errors where you encountering on `ROLLBACK TO SAVEPOINT cockroach_restart`?
And btw, the commit message is misleading; it talks about rollbacks, but `ROLLBACK TO SAVEPOINT cockroach_restart` isn't quite that.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw2xCRSv_5VXS95gOX1:bxku3zc)*

neeral:
@andreimatei
In our failure testing, we have our own logic to determine when to retry. For example, we will retry on ErrBadConnection because it's quite likely that was received because such a failure was injected into the cluster. We encountered examples where transactions would fail (call to `ExecuteInTx`) because it received an error when rolling back to a savepoint. Since the errors are being overwritten, it is difficult for us to handle these errors. We need to check for strings instead of error types. With this motivation, it seemed better to provide an error with more structure than just a string.

From reading the two linked issues, we're not using `github.com/pkg/errors` because it  'is arbitrary and non-standard'. I didn't understand if there was a more technical reason for not wrapping errors.

Do you have a suggestion on how to handle the use case we have?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw3jYKx35jS5yDrOz7S:b5wckhx)*

andreimatei:
It's not *you* whose use of `pkg/errors` is questionable. The CRDB codebase also uses that package, to great joy. However, `cockroach-go` is supposed to be used by the public at large who may not have agreed to add a dependency on that library.

If `driver.ErrBadConn` is the issue, I suggest you send a patch that doesn't "wrap" that particular error - similarly to what we do for `pq.Error` coming from `ROLLBACK TO SAVEPOINT cockroach_restart`(there's a comment in there that says that "the "driver: bad connection" is untyped", but maybe that's just wrong). Would that work?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw5RUkjVL5Tmt1mlRdJ:bt6ysf8)*

andreimatei:
On the other hand, if we were to stop wrapping "bad connection" errors, than I'm not sure there'd still be a point in wrapping any other errors (there shouldn't really be other errors).
Another alternative is to wrap everything (including `pq.Error`), but implement our own equivalent to `pkg/errors.Wrap()` that returns something with a `Cause()` method. This will make a `pkg/errors.Cause(crdbErr)` call in client code still work.

What do you think?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw5Zv_N6lWYl3H1Vteo:b1k6xf3)*

neeral:
`driver.ErrBadConn` is untyped in the sense that it is defined by `errors.New(<string>)`. Hence, it has no structure. Because of this, it doesn't have a `Message` field, we would have to overwrite the error, as per the comment.

This issue is not limited to `driver.ErrBadConn`: we've faced this issue with a small number (3-4) errors. Whether or not this list grows, handling each one individually in `cockroach-go` doesn't seem appropriate.

With regards to your second proposal, I think that could work. If I understand correctly, it suggests creating our own error struct and factory method that wraps errors. However, if we think that others (the public at large) aren't happy with using `github.com/pkg/errors`, will it be acceptable to them for us to create our own version?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw6Q-Rr1It0m7IYbsUW:b-qu1j0j)*

andreimatei:
I'm leaning towards saying that this library should get out of the business of wrapping anything. I think it's questionable whether it provides any benefit and it invites the present conversation.

> This issue is not limited to driver.ErrBadConn: we've faced this issue with a small number (3-4) errors. Whether or not this list grows, handling each one individually in cockroach-go doesn't seem appropriate.

What other errors have you encountered? Were they networking errors? I think the fact that you've encountered different types may be quite pertinent to the present conversation.

> However, if we think that others (the public at large) aren't happy with using github.com/pkg/errors, will it be acceptable to them for us to create our own version?

The fact that this lib would have its own "wrapped" error type would generally be transparent to clients. If the client is sophisticated enough to want to look at causes, it would be able to call `err.Cause()` directly or use `pkg/errors.Cause(err)` on it (if they're already using `pkg/errors`).

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw6Wkf4rjyIluIDgULm:b-9i2031)*

neeral:
The errors we have seen from `ExecuteInTx`, which aren't of type`pq.Error`, are: `driver.ErrBadConn`, `io.EOF` and `net.OpError`. I've only seen the ErrBadConn when rolling back to a savepoint.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kw6rvN0RznxJEL0GH7z:b-nyxu3p)*

andreimatei:
LGTM

Only nits, thanks for doing this.

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[crdb/error.go, line 5 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwArKGX-EpG_Q7uajs6:-KwArKGX-EpG_Q7uajs7:bg0dton) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/error.go#L5)):*
> ```Go
> import "fmt"
> 
> func newTxnRestartError(
> ```

nit: put the ctor (immediately) after the type definition

---

*[crdb/error.go, line 8 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAqoEpfbMFhV3teapr:-KwAqoEpfbMFhV3teaps:boz20uf) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/error.go#L8)):*
> ```Go
> 	err error, retryErr error, format string, args ...interface{},
> ) *TxnRestartError {
> 	if err == nil || retryErr == nil {
> ```

why the `retryErr == nil` test? I think that should either be an assertion (so, panic), or ignore it and return a `TxnRestartError` with a nil `RetryCause`.

---

*[crdb/error.go, line 20 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAthK0ujt9HstU6rKt:-KwAthK0ujt9HstU6rKu:bm2gkeb) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/error.go#L20)):*
> ```Go
> // TxnRestartError represents an error when restarting a transaction. It includes
> // the original error which triggered the restart.
> type TxnRestartError struct {
> ```

the two causes in this error are confusing enough; I think more words in the type comment explaining the distinction would be good

---

*[crdb/error.go, line 22 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAssWk1IwWoZbkz8_X:-KwAssWk1IwWoZbkz8_Y:bgumgxy) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/error.go#L22)):*
> ```Go
> type TxnRestartError struct {
> 	cause      error
> 	RetryCause error
> ```

since `cause` is unexported, I'd also make `RetryCause` unexported and give it an access method, for symmetry.

---

*[crdb/error.go, line 26 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwArP_jSSafqD9pm8Pu:-KwArP_jSSafqD9pm8Pv:b-fi6nrr) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/error.go#L26)):*
> ```Go
> }
> 
> func (w *TxnRestartError) Error() string { return w.msg + ": " + w.cause.Error() }
> ```

for crdb code, out linters enforce that all exported methods have comments that start with the method name. Let's keep this true for this lib too, please.
Say `// Error implements the error interface` and 
```
Cause() returns the error encountered by the "ROLLBACK TO SAVEPOINT cockroach_restart" statement.
This method also implements the internal pkg/errors.causer interface, so TxnRestartError works with pkg/errors.Cause().
```

---

*[crdb/tx.go, line 26 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAsBvy3u24JnVbc8SL:-KwAsBvy3u24JnVbc8SM:b-fhbsz6) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L26)):*
> ```Go
> )
> 
> // AmbiguousCommitError represents an error that left a transaction in an
> ```

move this to the new file and make it similar to `TxnRestartError` - give it a `Cause()` and a ctor.
If you want, you can use a common error type with a cause and embed it in the two exported errors.

---

*[crdb/tx.go, line 38 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAsaWiGWf2E_JiQj38:-KwAsaWiGWf2E_JiQj39:b-xz20os) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L38)):*
> ```Go
> // we err on RELEASE with a communication error it's unclear if the transaction
> // has been committed or not (similar to erroring on COMMIT in other databases).
> // In that case, we return AmbiguousCommitError.
> ```

please say here that TxnRestartError can be returned too

---

*[crdb/tx.go, line 43 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAtZcJ0r0CdOElTMYX:-KwAtZcJ0r0CdOElTMYY:b-tm5df5) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L43)):*
> ```Go
> // https://cockroachlabs.com/docs/stable/transactions.html.
> //
> // NOTE: the supplied exec closure should not have external side
> ```

not your code, but please take the opportunity to s/exec/fn.

---

*[crdb/tx.go, line 106 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAt9JuWOFmc6n0IfDM:-KwAt9JuWOFmc6n0IfDN:b-me2mjw) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L106)):*
> ```Go
> 		}
> 		if _, err = tx.ExecContext(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); err != nil {
> 			msgPattern := "restarting txn failed. ROLLBACK TO SAVEPOINT encountered error: %s. " +
> ```

this msg should be hardcoded in `TxnRestartError`, and `TxnRestartError.msg` should go away (iow, there's no reason for TxnRestartError.msg to be configurable and to burden the client with constructing a message).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-KwAttb22KjL43b8tdPZ:b-a0nese)*

neeral:
Thanks for the feedback @andreimatei. I've addressed the comments and think it's good to go now. If you're happy with it, please merge.

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[crdb/error.go, line 8 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAqoEpfbMFhV3teapr:-KwHBgamMKSAH6qpSCtU:bvm0rz9) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/error.go#L8)):*

why the `retryErr == nil` test? I think that should either be an assertion (so, panic), or ignore it and return a `TxnRestartError` with a nil `RetryCause`.
</blockquote></details>

I'll make it panic. Restarting a transaction if there was no error is an invalid state to reach.

---

*[crdb/tx.go, line 106 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAt9JuWOFmc6n0IfDM:-KwHMdenMbQxsRo7B4_v:b-9guop8) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L106)):*

this msg should be hardcoded in `TxnRestartError`, and `TxnRestartError.msg` should go away (iow, there's no reason for TxnRestartError.msg to be configurable and to burden the client with constructing a message).
</blockquote></details>

Good point. I also noticed that in the previous version, we were accessing `pqErr` and `pqErr.Code` without checking the value of `ok`. If the err wasn't a `*pq.Error`, this would have been a zero value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-KwHS5iCw8lXUERvZeBu:bpgw5pe)*

crdb/error.go
```diff
@@ -0,0 +1,60 @@
...
+func (e *txError) Cause() error { return e.cause }
+
+// AmbiguousCommitError represents an error that left a transaction in an
+// ambiguous state: unclear if it committed or not.
+type AmbiguousCommitError struct {
+	txError
+}
+
+func newAmbiguousCommitError(err error) *AmbiguousCommitError {
+	if err == nil {
```
tamird: all these nil checks are "more general" than necessary - it shouldn't be valid to pass nil values in any of these parameters. I'd just remove them all, but @andreimatei should weigh in.

andreimatei:
Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, all commit checks successful.

---

*[crdb/error.go, line 24 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-Kw7A-3ULzE3a2C6CF0W-r3-24:-Kw_gARJSbeMx996MdZf:b-8i82po) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/cdfe5cea4106125a7137482567dc0488e13a7139/crdb/error.go#L24)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

all these nil checks are "more general" than necessary - it shouldn't be valid to pass nil values in any of these parameters. I'd just remove them all, but @andreimatei should weigh in.
</blockquote></details>

yeah, I'd either remove or make it a panic like you did in the other ctor

---

*[crdb/tx.go, line 106 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAt9JuWOFmc6n0IfDM:-Kw_gyqnMucsD4QqYo66:b-ax5r06) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L106)):*
<details><summary><i>Previously, neeral (Neeral Dodhia) wrote…</i></summary><blockquote>

Good point. I also noticed that in the previous version, we were accessing `pqErr` and `pqErr.Code` without checking the value of `ok`. If the err wasn't a `*pq.Error`, this would have been a zero value.
</blockquote></details>

Isn't the order of the argument now backwards? The "retry error" is `pqErr` ( = err).

In the code before the change, I think `pqErr` was guaranteed to not be nil here, otherwise we'd have returned before. Notice `pqErr` vs `rollbackPQErr`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43)*

neeral:
Addressed andrei's and tamir's comments.

---

Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

*[crdb/error.go, line 24 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-Kw7A-3ULzE3a2C6CF0W-r3-24:-KwgI8NxGscKJJGjHVji:b-oicyv6) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/cdfe5cea4106125a7137482567dc0488e13a7139/crdb/error.go#L24)):*

yeah, I'd either remove or make it a panic like you did in the other ctor
</blockquote></details>

removed it.

---

*[crdb/tx.go, line 106 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-KwAt9JuWOFmc6n0IfDM:-KwgJ40eGrX05Hlnd-NH:b6zrrk4) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/bb1e148dec0b76c6683b0498a86efa39c3fd3f09/crdb/tx.go#L106)):*

Isn't the order of the argument now backwards? The "retry error" is `pqErr` ( = err).

In the code before the change, I think `pqErr` was guaranteed to not be nil here, otherwise we'd have returned before. Notice `pqErr` vs `rollbackPQErr`.
</blockquote></details>

Good catch!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-KwgJbiwAZ4Hfk60FuXm:b-3orw3h)*

crdb/error.go
```diff
@@ -0,0 +1,57 @@
...
+	txError
+	retryCause error
+	msg        string
+}
+
+func newTxnRestartError(err error, retryErr error) *TxnRestartError {
+	if retryErr != nil {
+		panic(fmt.Sprintf("no original error for TxnRestartError: %v", err))
+	}
+	if err == nil {
```
tamird: did you mean to remove this one?

neeral:
Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[crdb/error.go, line 40 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-Kw7A-3ULzE3a2C6CF0W-r4-40:-KwgLFq-h6hEu7f0Jp2z:b-7toitr) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/24dd465a99f258b715d724a224dd0022bf8572ff/crdb/error.go#L40)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

did you mean to remove this one?
</blockquote></details>

Forgot about this. Removed this one and the check on `retryErr` as well (it didn't make sense to check on one parameter and not the other)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43)*

neeral:
Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[crdb/error.go, line 47 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-Kwkhh7tqC6ZeFj0yp9O:-KwmAjYL9TneBSf-HLlx:b6zrrkh) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/1c55174afe7bd8e17568e0a4ed2f52ae40b23d50/crdb/error.go#L47)):*

I think you've got a recursive call here, no? I think you only want `return e.msg`.
</blockquote></details>

Good catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43)*

andreimatei:
Looks good, but there seems to be something wrong with the CI run. I've never seem this Travis CI before, I think it's new that we use it for this repo. I'll investigate on Monday.

---

Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kx-9ESqADB4paDCIgm6:bii3lpg)*

andreimatei:
@jordanlewis I have reason to believe you know something about the new TravisCI setup. It seems to run two jobs for this repo - "Go 1.x" and "Go 1.8.x". The latter seems to get stuck on a `go test -v ./...` invocation. Running that on my machine seems to hang too. If I ctrl-c it, it spews a bunch of:
```
2017/10/23 12:51:04 waiting for download of /var/folders/ft/76p8y6813bx1r_t48_qq5r400000gn/T/cockroach.darwin-amd64.054b6503ddee1bedd1081640059f027b952ec7a3
```
Does this mean anything to you?

---

Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kx97i9LYzAEAdoGDejo:bgxqggs)*

jordanlewis:
@andreimatei I'm afraid I don't know anything about Travis CI or why we're using it in this repo.

---

Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kx9Aagk5sQ8_f5qUHLI:b-gciq0v)*

tamird:
I'm pretty sure I set that up; there was no testing at all in this repo prior to that. I'll take a look at this failure.

andreimatei:
@neeral I've rebased and merged for you. Thanks!

---

Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/43#-:-Kx9cWjES952ZnXhaHuO:b-aotxx0)*

PR #44, opened 2017-10-18
travis: remove double go test

dvrkps:


tamird:
Thanks for the PR, but this is intentional. The first runs the tests without the race detector, the second runs them with it.

dvrkps:
@tamird what is reason ?

PR #45, opened 2017-10-23
Don't test packages in parallel 

tamird:
Along with some other fixes; see commits.


andreimatei:
LGTM; thanks bro.

Good catch! (I think; do you know what was going on exactly)?

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/45#-:-Kx9NFlw4bV6oILmA5jv:bskdudg)*

tamird:
I'm not totally sure, I wasn't able to reproduce this reliably on my system. There's probably still something to investigate here, but at least CI for the repository will work for now.

PR #46, opened 2017-12-27
crdb: detect wrapped HandledRetryableError

### LINKED ISSUE #16: should use `errors.Cause(err)` instead of `err` to check for retryable error

benesch:
Teach ExecuteInTx to detect when a retryable error has been wrapped with
errors.Wrap.

Currently, returning a wrapped error from an ExecuteInTx closure will
break automatic transaction retries, as the error is no longer directly
marked as retryable. This is particularly hard to notice when code
is split across multiple functions. Consider the following:

    func doSomething(ctx context.Context, tx Tx) error {
        if err := doSomethingElse(ctx, tx); err != nil {
            return errors.Wrap("doSomethingElse", err)
        }
    }

    ... 8< ...

    ExecuteInTx(ctx, tx, func() error {
        return doSomething(ctx, tx)
    })

If the actions taken by doSomethingElse cause a retryable txn error
to be returned, doSomething will wrap the error. ExecuteTx thus won't see
the retryable error and the operation will fail.

Counterpart of cockroachdb/cockroach#21068 and its associated issue,
cockroachdb/cockroach#17667.

Fix #16.

### DIFF
--- a/crdb/error.go
+++ b/crdb/error.go
@@ -2,16 +2,39 @@ package crdb
 
 import "fmt"
 
+// ErrorCauser is the type implemented by an error that remembers its cause.
+//
+// ErrorCauser is intentionally equivalent to the causer interface used by
+// the github.com/pkg/errors package.
+type ErrorCauser interface {
+	// Cause returns the proximate cause of this error.
+	Cause() error
+}
+
+// errorCause returns the original cause of the error, if possible. An error has
+// a proximate cause if it implements ErrorCauser; the original cause is the
+// first error in the cause chain that does not implement ErrorCauser.
+//
+// errorCause is intentionally equivalent to pkg/errors.Cause.
+func errorCause(err error) error {
+	for err != nil {
+		cause, ok := err.(ErrorCauser)
+		if !ok {
+			break
+		}
+		err = cause.Cause()
+	}
+	return err
+}
+
 type txError struct {
 	cause error
 }
 
-// Error implements the error interface
+// Error implements the error interface.
 func (e *txError) Error() string { return e.cause.Error() }
 
-// Cause returns the error encountered by the "ROLLBACK TO SAVEPOINT
-// cockroach_restart" statement. This method also implements the internal
-// pkg/errors.causer interface, so TxnRestartError works with pkg/errors.Cause().
+// Cause implements the ErrorCauser interface.
 func (e *txError) Cause() error { return e.cause }
 
 // AmbiguousCommitError represents an error that left a transaction in an
@@ -43,9 +66,8 @@ func newTxnRestartError(err error, retryErr error) *TxnRestartError {
 	}
 }
 
-// Error implements the error interface
+// Error implements the error interface.
 func (e *TxnRestartError) Error() string { return e.msg }
 
-// RetryCause returns the error encountered by the transaction, which caused the
-// transaction to be restarted.
+// RetryCause returns the error that caused the transaction to be restarted.
 func (e *TxnRestartError) RetryCause() error { return e.retryCause }
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -38,6 +38,32 @@ import (
 //
 // NOTE: the supplied fn closure should not have external side
 // effects beyond changes to the database.
+//
+// fn must take care when wrapping errors returned from the database driver with
+// additional context. For example, if the UPDATE statement fails in the
+// following snippet, the original retryable error will be masked by the call to
+// fmt.Errorf, and the transaction will not be automatically retried.
+//
+//    crdb.ExecuteTx(ctx, db, txopts, func (tx *sql.Tx) error {
+//        if err := tx.ExecContext(ctx, "UPDATE..."); err != nil {
+//            return fmt.Errorf("updating record: %s", err)
+//        }
+//        return nil
+//    })
+//
+// Instead, add context by returning an error that implements the ErrorCauser
+// interface. Either create a custom error type that implements ErrorCauser or
+// use a helper function that does so automatically, like pkg/errors.Wrap:
+//
+//    import "github.com/pkg/errors"
+//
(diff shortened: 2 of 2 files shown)


andreimatei:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[crdb/cause.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-L1OL8o9E8YHjFLFtObE:-L1OL8oA6y5B9v8-I15H:b9vrbhy) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/01adbcf6485f9a1177a127961b9160028ffe4971/crdb/cause.go#L1)):*
> ```Go
> // Copyright (c) 2015, Dave Cheney <dave@cheney.net>
> ```

there's already errors.go, which mentions the `pkg/errors.Causer` interface. I think this code should be around there.

---

*[crdb/tx.go, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-L1OLNkRw-fPGu-rIVti:-L1OLNkRw-fPGu-rIVtj:bmzlrad) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/01adbcf6485f9a1177a127961b9160028ffe4971/crdb/tx.go#L62)):*
> ```Go
> // closure (fn). Any statements performed on the tx before ExecuteInTx is invoked will *not*
> // be re-run if the transaction needs to be retried.
> func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
> ```

we should comment that `fn` can use pkg/errors to do wrapping

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-:-L1OLSOeJvJgI99bQlHc:bnfp4nl)*

bdarnell:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[crdb/cause.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-L1OL8o9E8YHjFLFtObE:-L1Yo-ifOmpY4ZBYC9EI:b-afldm4) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/01adbcf6485f9a1177a127961b9160028ffe4971/crdb/cause.go#L1)):*

there's already errors.go, which mentions the `pkg/errors.Causer` interface. I think this code should be around there.
</blockquote></details>

Well, it's got a separate license, so from that perspective it's better to have it in a separate file (although to be fully correct we need to note this in the top level as we did in the main repo). Or we could decide that the code here is trivial enough that we can include it without bringing in all the legalese. 

---

*[crdb/tx.go, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-L1OLNkRw-fPGu-rIVti:-L1YpVIEAg78AYWagxtr:bug51yv) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/01adbcf6485f9a1177a127961b9160028ffe4971/crdb/tx.go#L62)):*

we should comment that `fn` can use pkg/errors to do wrapping
</blockquote></details>

Yeah, we should mention both `pkg/errors` and the fact if you wrap errors in other ways you can implement `Cause`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-:-L1Ypn_kQ0Hwv1uF3Ky5:bnfp4nl)*

benesch:
Will do on the doc update. Ben, do you have a preference for which way to
go with the licensing?

bdarnell:
I'd just copy it without the license. 

Also, make the Causer interface public so you can refer to it in docs (instead of an inline copy in both the code and docs).

benesch:
Done. PTAL!

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[crdb/tx.go, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-L1OLNkRw-fPGu-rIVti:-L1s-8mEUo-ukT9vj1OA:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/01adbcf6485f9a1177a127961b9160028ffe4971/crdb/tx.go#L62)):*
<details><summary><i>Previously, bdarnell (Ben Darnell) wrote…</i></summary><blockquote>

Yeah, we should mention both `pkg/errors` and the fact if you wrap errors in other ways you can implement `Cause`. 
</blockquote></details>

Done.

---

*[crdb/cause.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-L1OL8o9E8YHjFLFtObE:-L1s-Birfm9SuL3uvNrt:bm3y2aw) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/01adbcf6485f9a1177a127961b9160028ffe4971/crdb/cause.go#L1)):*
<details><summary><i>Previously, bdarnell (Ben Darnell) wrote…</i></summary><blockquote>

Well, it's got a separate license, so from that perspective it's better to have it in a separate file (although to be fully correct we need to note this in the top level as we did in the main repo). Or we could decide that the code here is trivial enough that we can include it without bringing in all the legalese. 
</blockquote></details>

Ok, I just copied the code directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-:-L1s-GpiOGpZ9TODKpr-:b-qsvnon)*

danhhz:
:lgtm_strong:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach-go/46#-:-L59g5K9JdanZDiScb_h:b-jbsf8z)*

PR #49, opened 2018-09-28
support pgx errors

### LINKED ISSUE #48: pgx support
My issue is connected to #16.
I faced the issue that none of our transactions was retried.
During the investigation, we realized, that in this commit was added the dependency to lib/pq and it handles only errors arisen by pq driver, but we use jackc/pgx. I think we all know, that it's quite popular.
And it isn't clear from [docs](https://www.cockroachlabs.com/docs/stable/build-a-go-app-with-cockroachdb.html#transaction-with-retry-logic)  that there is support only for lib/pq.

Maybe we could find the solution which will be driver independent?

RaduBerinde:
This change adds code to recognize `pgx` errors as well. Note that
unlike `lib/pq`, merely importing `pgx` doesn't register a `sql`
driver (that's done by `pgx/stdlib`).

Fixes #48.

### DIFF
--- a/crdb/README.md
+++ b/crdb/README.md
@@ -1,0 +1,9 @@
+CRDB
+====
+
+`crdb` is a wrapper around the logic for issuing SQL transactions which performs
+retries (as required by CockroachDB).
+
+Note that unfortunately there is no generic way of extracting a pg error code;
+the library has to recognize driver-dependent error types. We currently support
+`github.com/lib/pq` and `github.com/jackc/pgx`.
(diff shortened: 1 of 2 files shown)


RaduBerinde:
Would they really care? I assume most people would just do a `go get` which will get the dependencies.

benesch:
> That's why I was hesitating in doing this, in the hope that we can get the drivers to expose a common error interface.

Is there an upstream issue to this effect?

andreimatei:
> Is there an upstream issue to this effect?

No.

Since I'm not doing anything about this myself, I'm going to get off the can.
LGTM

PR #50, opened 2018-10-09
add variant of ExecuteTx that runs on a specific sql.Conn

RaduBerinde:
I considered making an interface for `BeginTx` so you can pass either `*sql.DB` or `*sql.Conn` but it ended up being more hassle documenting what's going on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: go
 
 go:
 - 1.x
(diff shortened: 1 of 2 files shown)


RaduBerinde:
Hm, for some reason I didn't realize ExecuteInTx is exported and I can just use that. Closing.

PR #53, opened 2019-09-12
support unwrapping of go1.13 errors

### LINKED ISSUE #52: Support Go 1.13 error wrapping in ExecuteTx
Go 1.13 has built in support for error wrapping and unwrapping: https://golang.org/pkg/errors/
There is a transitional package at https://godoc.org/golang.org/x/xerrors

Problem: When using `ExecuteTx`, errors returned from the user supplied function must [wrap the underlying error correctly](https://godoc.org/github.com/cockroachdb/cockroach-go/crdb#ExecuteTx), so that the transaction can be retried. This worked well previously as long as you implemented the `ErrorCauser` interface or use `pkg/errors` to wrap the underlying error.

In Go 1.13, to wrap errors, there is a `%w` placeholder: `fmt.Errorf("error doing something: %w", err)`. You can then use `errors.Unwrap()` to retrieve the underlying error, use `Is` to assert the error type or use `As` to retrieve the error if it matches a type.

I think this package should be updated to support Go 1.13 error wrapping in addition to the current `ErrorCauser` interface as error wrapping using the stdlib's `errors` package is the recommended way, moving forward. We are currently in the process of migrating to Go 1.13 and are dropping `pkg/errors` for stdlib errors where possible.

andreimatei:
The library currently supports handling errors wrapped by pkg/errors
inside the txn closure. This patch adds similar support for go1.13
"errors" package.

Fixes #52

### DIFF
--- a/crdb/error.go
+++ b/crdb/error.go
@@ -11,18 +11,46 @@ type ErrorCauser interface {
 	Cause() error
 }
 
+// UnwrappableError describes errors compatible with errors.Unwrap.
+type UnwrappableError interface {
+	// Unwrap returns the proximate cause of this error.
+	Unwrap() error
+}
+
+// Unwrap is equivalent to errors.Unwrap. It's implemented here to maintain
+// compatibility with Go versions before 1.13 (when the errors package was
+// introduced).
+// It returns the result of calling the Unwrap method on err, if err's type
+// implements UnwrappableError.
+// Otherwise, Unwrap returns nil.
+func Unwrap(err error) error {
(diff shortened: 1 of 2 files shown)


knz:
Why don't you import cockroachdb/errors and simply use errors.UnwrapOnce/UnwrapAll, which automatically supports both Cause() and Unwrap()?

I think it would simplify. (And maybe there are a few other simplifications available too to the rest of that code)
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

knz:
ok I've done the rebase for you

Issue #60, opened 2020-03-29
pgx compatibility


georgysavva:
In my previous [issue](https://github.com/cockroachdb/cockroach-go/issues/59) I was advised to use CockroachDB with the [pgx](https://github.com/jackc/pgx) library. And I started to look deeper into it.
I found out that pgx.Tx isn't compatible with the crdb.Tx interface and I can't use the crdb.ExecuteInTx() function.
I found a wrapper in the main cockroach repo: [here](https://github.com/cockroachdb/cockroach/blob/92495ffb2182671b3338a8c6cd2734cb6de5d392/pkg/workload/pgx_helpers.go#L162-L183)
But it uses very old version of the library, you can see it in the import statement: 
import "github.com/jackc/pgx"
and the current import path is "github.com/jackc/pgx/v4" 
The wrapper wasn't updated in 2 years and calls methods that no longer exist in pgx.

Aside from this incompatibility issue, wouldn't it be better to put those helpers in this repo, instead of the main server repo?

andreimatei:
We used to need to depend on a particular pgx version because of https://github.com/jackc/pgconn/issues/15. But now improvements have been made there, so I think we can now deal with a couple of versions without importing any one.

> Aside from this incompatibility issue, wouldn't it be better to put those helpers in this repo, instead of the main server repo?

Probably. However you can still use pgx with our library in its "driver" mode - so, using pgx through Go's `database/sql` package, not directly (modulo the version comment above). `cockroach-go` is mostly aimed at working with `database/sql`, although its mandate could be extended.

georgysavva:
Thanks for the advise with `database/sql`. I checked how `pgx` integrates with `sql` library and it seems, that `pgx` implementation of the `driver.Tx` doesn't have access to the context on `driver.Tx.Commit()` and `Rollback()`. I opened an issue about that: https://github.com/jackc/pgx/issues/703
It seems we will be able to fix it pretty easily, hence it's not a big deal. But the general advice of the `pgx` maintainer is to use `pgx` natively (not "driver" mode) whenever possible. So supporting last version of `pgx` would be a desirable feature. Thanks!

PR #61, opened 2020-04-05
Make crdb compatible with pgx.Tx

georgysavva:
Hi. As I was mentioning in this issue https://github.com/cockroachdb/cockroach-go/issues/60, crdb package doesn't support native pgx.Tx and we concluded that it would be a good feature to add. It turned out that it's not so difficult to do (or I missing something) and I decided to do it on my own, if you don't mind)

### DIFF
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -21,6 +21,7 @@ import (
 	"database/sql"
 
 	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx"
 	"github.com/lib/pq"
 )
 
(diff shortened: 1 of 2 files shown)


andreimatei:
Hey Georgy! Thanks for working on this stuff!
I've got some thoughts on this PR as written, and generally this library. I'll respond tomorrow.

andreimatei:
@georgysavva please take a look at #62 instead of this.

I've created a dedicated sub-package for pgx support.
There were two things that bothered me with your PR:
- adding a dependency on pgx from crdb
- the `ExecuteInTx` API is inherently dangerous because it assumes that the transaction had not previously been use. If it had, in fact, been used, you're in for bad surprises. `ExecuteTx` is the better API, as it controls the creation of the transaction.

If you like that patch, let's close this one. Sorry for not using your code; I thought it's easier if I just do it since I wanted to go to town a little bit.

georgysavva:
I don't mind at all, your patch looks great, closing this one.

PR #62, opened 2020-04-06
add support for pgx and gorm 

### LINKED ISSUE #60: pgx compatibility

andreimatei:
Before this patch, the library only had support for Go's sql package.
This patch adds subpackages for pgx and gorm, providing suitable
ExecuteTx() functions. We had a function called ExecuteInTx which was
intended to be used with transactions initiated by other frameworks.
However the function's abstraction of a Txn did not match any framework,
and the API was generally dangerous because it simply assumed that the
respective txn had not previously been used to run statements.
In contrast, the new ExecuteTx() functions work with the respective
frameworks and controls the creation of the transaction.

This patch also makes the crdb package not import a pgx package any
more.

Fixes #60

### DIFF
--- a/crdb/crdbpgx/pgx.go
+++ b/crdb/crdbpgx/pgx.go
@@ -1,0 +1,65 @@
+// Copyright 2016 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package crdbpgx
+
+import (
+	"context"
+	"github.com/cockroachdb/cockroach-go/crdb"
+
+	"github.com/jackc/pgx"
+)
+
+// ExecuteTx runs fn inside a transaction and retries it as needed. On
+// non-retryable failures, the transaction is aborted and rolled back; on
+// success, the transaction is committed.
+//
+// See crdb.ExecuteTx() for more information.
+//
+// conn can be a pgx.Conn or a pgxpool.Pool.
+func ExecuteTx(
+	ctx context.Context, conn Conn, txOptions pgx.TxOptions, fn func(pgx.Tx) error,
+) error {
+	tx, err := conn.BeginTx(ctx, txOptions)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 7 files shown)


andreimatei:
cc @georgysavva

crdb/crdbpgx/pgx.go
```diff
@@ -0,0 +1,53 @@
...
+	ctx context.Context, conn Conn, txOptions pgx.TxOptions, fn func(pgx.Tx) error,
+) error {
+	tx, err := conn.BeginTx(ctx, txOptions)
+	if err != nil {
+		return err
+	}
+	return crdb.ExecuteInTx(ctx, pgxTxAdapter{tx}, func() error { return fn(tx) })
+}
+
+// Conn abstracts pgx transactions creators: pgx.Conn and pgxpool.Connect.
```
georgysavva, line 37: `pgxpool.Connect` is a function. It should be pgxpool.Pool?

crdb/crdbpgx/pgx.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/jackc/pgx"
+)
+
+// ExecuteTx runs fn inside a transaction and retries it as needed. On
+// non-retryable failures, the transaction is aborted and rolled back; on
+// success, the transaction is committed.
+//
+// See crdb.ExecuteTx() for more information.
+//
+// conn can be a pgx.Conn or a pgxpool.Conn.
```
georgysavva, line 26: or a pgxpool.Pool

georgysavva:
Looks great, thanks for implementing this! 
I have one more question:
In the old `pgx_helpers.go` file you had this [MultiConnPool](https://github.com/cockroachdb/cockroach/blob/6365a3f21e467f3321068fa983d3999415a42609/pkg/workload/pgx_helpers.go#L52) struct. That creates a `pgx.Pool` for each URL. Does it make sense if all cockroachDB nodes are behind the cloud global TCP load balancer that gives a single anycast IP and hence single URL?

georgysavva:
Thanks. Got it.

PR #75, opened 2020-06-12
testserver: run cockroach server in secure mode

asubiotto:
The server would previously run in insecure mode. This change makes the test
environment more in line with what a production environment would be like.

This change is transparent to the client as the PGURL used is returned through
a TestServer level method, which now includes secure connection options.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -99,8 +99,8 @@ type TestServer struct {
 // opens a SQL database connection to it. Returns a sql *DB instance a
 // shutdown function. The caller is responsible for executing the
 // returned shutdown function on exit.
-func NewDBForTest(t *testing.T) (*sql.DB, func()) {
-	return NewDBForTestWithDatabase(t, "")
+func NewDBForTest(t *testing.T, opts ...testServerOpt) (*sql.DB, func()) {
+	return NewDBForTestWithDatabase(t, "", opts...)
 }
 
 // NewDBForTestWithDatabase creates a new CockroachDB TestServer
@@ -108,8 +108,8 @@ func NewDBForTest(t *testing.T) (*sql.DB, func()) {
 // specified, the returned connection will explicitly connect to
 // it. Returns a sql *DB instance a shutdown function. The caller is
 // responsible for executing the returned shutdown function on exit.
-func NewDBForTestWithDatabase(t *testing.T, database string) (*sql.DB, func()) {
-	ts, err := NewTestServer()
+func NewDBForTestWithDatabase(t *testing.T, database string, opts ...testServerOpt) (*sql.DB, func()) {
+	ts, err := NewTestServer(opts...)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -141,11 +141,30 @@ func NewDBForTestWithDatabase(t *testing.T, database string) (*sql.DB, func()) {
 	}
 }
 
+type testServerOpt func(args *testServerArgs)
+
+type testServerArgs struct {
+	secure bool
+}
+
+// SecureOpt is a TestServer option that can be passed to NewTestServer to
+// enable secure mode.
+func SecureOpt() testServerOpt {
+	return func(args *testServerArgs) {
+		args.secure = true
+	}
+}
+
 // NewTestServer creates a new TestServer, but does not start it.
 // The cockroach binary for your OS and ARCH is downloaded automatically.
 // If the download fails, we attempt just call "cockroach", hoping it is
 // found in your path.
-func NewTestServer() (*TestServer, error) {
+func NewTestServer(opts ...testServerOpt) (*TestServer, error) {
+	serverArgs := &testServerArgs{}
+	for _, applyOptToArgs := range opts {
+		applyOptToArgs(serverArgs)
+	}
+
 	var cockroachBinary string
 	var err error
 	if len(*customBinary) > 0 {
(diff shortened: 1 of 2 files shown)


asubiotto:
The only thing to be wary of is that a caller using a custom pgurl will need to change its behavior. This might justify the introduction of parameters to make insecure optional but I haven't seen a callsite where this is true yet.

asubiotto:
The motivation is to mimic a production environment in `TestServer` as much as we can. The idea is to add tenant creation to this package too and use that in examples-orms so that we can start running those tests in a multitenant environment.

Adding certs doesn't cause any noticeable performance difference. `NewTestServer` downloads a cockroach binary, which most likely be the majority of overhead in any test that uses this.

> Another thing is that I think we should try to make this TestServer be in-memory only
This package is currently written in order to have an easy way to spin up a full cockroach server, this change is in line with that goal. If we want to add this in-memory functionality, I would prefer have tests choose whether they want to use on-disk or in-memory storage rather than replacing the current code. In any case, I think this is beyond the scope of this PR and is an argument more about the purpose of this package.

andreimatei:
I'm not gonna comment any more,  but I think going to secure by default is a bad idea. You've also said something about multi-tenants; I'm not sure what that entails, but I think that if it entails any more changes to this library you should consider more if this is the right code to use at all.

> The motivation is to mimic a production environment in TestServer as much as we can. The idea is to add tenant creation to this package too and use that in examples-orms so that we can start running those tests in a multitenant environment.

I don't think you can just say "mimic production" and leave it as that. This library is meant for people's unit tests. Obviously everybody's production environment is gonna be very different from what this library gives you. What's important is for the functional behavior to be CRDB's when it comes to query processing. Similarity to production for other operational concerns is not a particular goal, I don't think. To the contrary, what matters more is having a small footprint, being easy to use, having as little magic as possible.
By creating certificates on disk (by default) you're increasing the footprint and the magic. One can no longer put a wireshark to debug a test. One can no longer easily create connection string if the one returned by `PGURL()` is not sufficient for whatever reason (I think the library already doesn't let you build a url by hand, because it's a bad library. But this patch is making it worse.)

> Adding certs doesn't cause any noticeable performance difference. NewTestServer downloads a cockroach binary, which most likely be the majority of overhead in any test that uses this.

I'm talking about query processing speed, not startup cost.

> This package is currently written in order to have an easy way to spin up a full cockroach server, this change is in line with that goal. 

The goal is to be able to run tests again CRDB. Don't read anything more into this code.

> If we want to add this in-memory functionality, I would prefer have tests choose whether they want to use on-disk or in-memory storage rather than replacing the current code. In any case, I think this is beyond the scope of this PR and is an argument more about the purpose of this package.

What I'm saying is that this patch is going in a wrong direction. Its scope doesn't matter.

asubiotto:
Modified this patch to use `--insecure` by default with the ability to pass in an option to run secure mode. This makes all current uses of `NewTestServer` insecure, as it was before.

> You've also said something about multi-tenants; I'm not sure what that entails, but I think that if it entails any more changes to this library you should consider more if this is the right code to use at all.

We want to verify that simple ORM interactions work in an end-to-end multitenant environment by using this package in `examples-orms`. It's possible that we're trying to do too much. @tbg what do you think?

I think we both agree that this package is supposed to provide an easy way to test against a cockroach server, but we fundamentally disagree on the scope. Part of the problem is that the purpose of this package isn't clearly stated anywhere, so there is no alternative but read into the code. This package already does plenty of involved magic in order to achieve this, which makes me believe that this magic is the value proposition of this package. I think ultimately, the package owner should decide on what the scope is. @rafiss I think this is you, could you weigh in?

tbg:
I agree that we should leave the defaults unchanged to avoid breaking existing users of this package, but other than that I don't quite understand @andreimatei's point. This package exists to make testing easier. Testing secure connections is something you may want to do and greasing that is not a step in the wrong direction as far as I can tell. That we're adding tenancy to it (down the road) can be argued about as a matter of style, but note that our main cockroach binary does expose tenancy functionality for the same reason I think it makes sense here -- keeping things simple. There's just no need to make it more complicated by maintaining two pieces of code when one will always subsume the other.

Ultimately, I would say this is @rafiss' call. I know [he's fine with TLS for examples-ORMs](https://github.com/cockroachdb/cockroach/issues/49109#issuecomment-632083285), but possibly there's a reservation against changing `cockroach-go` - I wouldn't assume so but let's see. @rafiss?

rafiss:
I'm in favor of making it easier to test with secure mode. I'd also lean towards making secure mode the default, but I agree that if there's a concern of too much fallout from breaking existing tests, then we don't need to change the default now.

My argument in favor of secure mode is similar to what's been said above: this library is for testing with CockroachDB end-to-end. We should make it as easy as possible to test in the same way that users are going to use the database in reality. Sure, it's going to differ from what people have in production, but I think "security" is a sufficiently important concept that it deserves to have first-class support in a testing library, and should be a high priority even when writing basic tests. I say this with the premise that security-related features are perennially under-tested in many software projects including CockroachDB. (And I don't mean anything fancy -- even basic things like "can we connect" have become broken due to lack of testing. See #48294 for an example with Java; something similar surely could conceivably happen with Go.) I'm not going to dispute that a small amount of pain will be added by testing in secure mode, but I think it's worth accepting that price for the chance of catching more show-stopping bugs.

The following analogy may be too contrived, but imagine an alternate universe where CockroachDB supported multiple isolation levels. We recommend everyone to use SERIALIZABLE, but the initial implementation of this testing library uses something lower because it makes the tests run faster and also tests that use transactions won't have to worry so much about retrying. After all, adding retry logic makes the footprint bigger, and adds more behind-the-scenes magic To me, that makes about as much sense as intentionally running everything in insecure mode because it's faster.

Perhaps the difference in opinion comes down to answering this question:
> What's important is for the functional behavior to be CRDB's when it comes to query processing.

I agree, and I would view as connecting in secure mode as a fundamental part of "query processing."

andreimatei:
> I agree that we should leave the defaults unchanged to avoid breaking existing users of this package

> I'd also lean towards making secure mode the default, but I agree that if there's a concern of too much fallout from breaking existing tests, then we don't need to change the default now.

I don't think we can break anybody, can we? There was no way to construct a URL by hand since you can't get your hands on the server's port, can you?
So if y'all love secure mode so much, I don't think that's the argument for not making it the default.

FWIW, I only buy the https://github.com/cockroachdb/docs/issues/7258 argument to a point. We're pretending like we're having the client test interact with a secure cluster, but we're generating the certificates for them. We're not actually asking the client to go through the steps normally required for producing and using a certificate so, for example, if another TLS change comes along and causes some client certs to not be accepted by the server, then these tests wouldn't tell you anything about it.

PR #78, opened 2020-06-29
Merge version-module branch

### LINKED ISSUE #73: Sync changes from master to version-module branch
Recently added `sqlx.Tx` adapter support to `crdb.Tx`, ref: https://github.com/cockroachdb/cockroach-go/commit/0b2acdb6dcb5095fa831fa048594a6e3a28bcd85 - What is the best way to get this change ported over to `version-module` branch and cut a release (say v2.1.0 or something)?

cc @andreimatei @rafiss

rafiss:
Now that cockroachdb/cockroach uses modules itself, this library
should support modules as a first-class.

fixes #73


rafiss:
@asubiotto I'm tagging you since I'm hoping to make sure I correctly handled the new tenant code in this merge.

testserver/testserver_test.go
```diff
@@ -91,14 +88,11 @@ func newTenantDBForTest(t *testing.T, secure bool) (*sql.DB, func()) {
...
 	}
-	if err := ts.Start(); err != nil {
-		t.Fatal(err)
-	}
 	tenant, err := ts.(tenantInterface).NewTenantServer()
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := tenant.Start(); err != nil {
+	if err := tenant.WaitForInit(); err != nil {
```
asubiotto, line 95: After changing the `ts.WaitForInit` call in `NewTenantServer` to `tenant.WaitForInit` you should be able to remove this.

testserver/tenant.go
```diff
@@ -124,5 +123,11 @@ func (ts *testServerImpl) NewTenantServer() (TestServer, error) {
 	tenantURL := *pgURL
 	tenantURL.Host = sqlAddr
 	tenant.setPGURL(&tenantURL)
+	if err := tenant.Start(); err != nil {
+        return nil, err
+	}
+	if err := ts.WaitForInit(); err != nil {
```
asubiotto, line 129: I think this should be `tenant.WaitForInit`

testserver/tenant.go
```diff
@@ -74,7 +73,7 @@ func (ts *testServerImpl) NewTenantServer() (TestServer, error) {
 		return nil, err
 	}
 	defer db.Close()
-	if err := ts.WaitForInit(db); err != nil {
+	if err := ts.WaitForInit(); err != nil {
```
asubiotto, line 76: Might make sense to order this before the `sql.Open` call

testserver/testserver.go
```diff
@@ -311,8 +315,13 @@ func (ts *testServerImpl) setPGURL(u *url.URL) {
 }
 
 // WaitForInit retries until a connection is successfully established.
-func (ts *testServerImpl) WaitForInit(db *sql.DB) error {
+func (ts *testServerImpl) WaitForInit() error {
```
asubiotto, line 318: Slight nit that doesn't need to be fixed in this PR, but it makes more sense for this method to retain a handle to a `db` rather than one to a `testServerImpl` since it's not really used. Maybe this should be a package-level function that takes in a `db` rather than a method on `testServerImpl`.

rafiss:
This is ready for another look

testserver/testserver.go
```diff
@@ -282,6 +273,19 @@ func NewTestServer(opts ...testServerOpt) (TestServer, error) {
 		curTenantID:      firstTenantID,
 	}
 	ts.pgURL.set = make(chan struct{})
+
+	if err := ts.Start(); err != nil {
+		return nil, err
+	}
+
+	if ts.PGURL() == nil {
```
asubiotto, line 281: I don't think this call is necessary (we set the `pgurl` above) but feel free to keep it in.

crdb/crdbgorm/gorm.go
```diff
@@ -17,7 +17,7 @@ package crdbgorm
 import (
 	"context"
 	"database/sql"
-	"github.com/cockroachdb/cockroach-go/crdb"
+	"github.com/cockroachdb/cockroach-go/v2/crdb"
```
otan, line 20: any reasons to upgrade this to v2? we should really only be doing this if the API is compatible with before. v1 should be fine.

go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/cockroachdb/cockroach-go/v2
+
+go 1.13
```
otan, line 3: any reason this is not 1.14?

crdb/crdbgorm/gorm.go
rafiss: If I remove the `v2`, then when I run the tests, these lines get added to `go.mod` and `go.sum`. is that bad?

```
diff --git a/go.mod b/go.mod
index 4945e3f..b73b6c9 100644
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/cockroachdb/cockroach-go/v2
 go 1.13

 require (
+    github.com/cockroachdb/cockroach-go v2.0.1+incompatible
     github.com/jackc/pgx/v4 v4.6.0
     github.com/jinzhu/gorm v1.9.12
     github.com/jmoiron/sqlx v1.2.0
diff --git a/go.sum b/go.sum
index 3f24f28..931b93a 100644
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,7 @@
 github.com/cockroachdb/apd v1.1.0 h1:3LFP3629v+1aKXU5Q37mxmRxX/pIu1nijXydLShEq5I=
 github.com/cockroachdb/apd v1.1.0/go.mod h1:8Sl8LxpKi29FqWXR16WEFZRNSz3SoPzUzeMeY4+DwBQ=
+github.com/cockroachdb/cockroach-go v2.0.1+incompatible h1:rkk9T7FViadPOz28xQ68o18jBSpyShru0mayVumxqYA=
+github.com/cockroachdb/cockroach-go v2.0.1+incompatible/go.mod h1:XGLbWH/ujMcbPbhZq52Nv6UrCghb1yGn//133kEsvDk=
 github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f h1:JOrtw2xFKzlg+cbHpyrpLDmnN1HqhBfnX7WDiW7eG2c=
 github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
```

go.mod
rafiss: no reason -- that's just what was getting merged from the `version-module` branch, and the file was created a while ago. should i change to 1.14 as part of this PR?

rafiss:
@otan thanks for the review! curious if you had any thoughts on Andrei's questions above?

https://reviewable.io/reviews/cockroachdb/cockroach-go/78#-MBZoPTY5z6OC4iSANon:-MB_2vhmDuPdk5sxvPkY:brjrh6h
and
https://reviewable.io/reviews/cockroachdb/cockroach-go/78#-MBZqCOaCHWUlc81pjig:-MB_MLPg5hBYwdv1jFDp:bm3vpxa

crdb/crdbgorm/gorm.go
otan: change the line `module github.com/cockroachdb/cockroach-go/v2` to `module github.com/cockroachdb/cockroach-go`

go.mod
otan: don't think it matters too much, either way.

crdb/crdbgorm/gorm.go
rafiss: my concern with that is that v2 was already created in this commit: https://github.com/cockroachdb/cockroach-go/commit/499e2dd981f98ec81e74d5888e06a7b30508edf1

and some people have been using the v2 version already, since the `version-module` branch was published here

it seems like bad things will happen if we make a new release that has a smaller version number than what people are using already. but maybe it's fine?

otan: if it's published, then yeah make sense to upgrade to v2.

rafiss:
thanks for all the reviews everyone! I'm going to go ahead and merge. wish me luck

otan:
___
*[go.mod, line 1 at r7](https://reviewable.io/reviews/cockroachdb/cockroach-go/78#-MBZoPTY5z6OC4iSANon:-MBfE95o5D_SNWB0f8uW:b-ldpdrd) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/39d9a5985bc02ec6c920d9c3297f937ade8d95a9/go.mod#L1)):*

i tried some stuff and from what i can tell, `import github.com/cockroachdb/cockroach-go` without the v2 does still work
</blockquote></details>

it does but imports the old version i think (observational, could be wrong!)

otan:
___
*[go.mod, line 1 at r7](https://reviewable.io/reviews/cockroachdb/cockroach-go/78#-MBZoPTY5z6OC4iSANon:-MBfo0JlAUBNupPoZVk5:b-aeayeh) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/39d9a5985bc02ec6c920d9c3297f937ade8d95a9/go.mod#L1)):*

> if you import github.com/cockroachdb/cockroach-go/crdb you get this commit: https://github.com/cockroachdb/cockroach-go/tree/e17c5b8c131ecc729416b61c0e0ec4114e1a4382

Do you understand why? I don't see any tags pointing to it.
</blockquote></details>

go mod automagically figures out the first non v2+ version.

PR #81, opened 2020-07-10
testserver: no longer rely on auto-init

irfansharif:
`cockroach start` without `--join` is deprecated behavior, and something
we're looking to get rid of in cockroachdb/cockroach#51245.


irfansharif:
Thanks!

PR #82, opened 2020-07-23
Fix start-single-node error with v19.1

rafiss:
The testserver package was recently changed to always use
start-single-node. However, v19.1 does not have this subcommand. To work
around this, we just try using start if start-single-node hits an error.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -277,7 +277,29 @@ func NewTestServer(opts ...testServerOpt) (TestServer, error) {
 	ts.pgURL.set = make(chan struct{})
 
 	if err := ts.Start(); err != nil {
-		return nil, err
+        // v19.1 and earlier do not have the `start-single-node` subcommand,
+        // so retry with `start`.
+        // TODO(rafi): Remove the `start` once we stop testing 19.1.
+        if !strings.Contains(ts.Stderr(), "unknown command \"start-single-node\"") {
+            return nil, err
+        }
+        log.Printf("Retrying after error with start-single-node: %s", err)
+        args[1] = "start"
+        ts = &testServerImpl{
+            serverArgs:       *serverArgs,
+            state:            stateNew,
+            baseDir:          baseDir,
+            cmdArgs:          args,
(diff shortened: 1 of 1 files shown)


testserver/testserver.go
```diff
@@ -277,7 +277,36 @@ func NewTestServer(opts ...testServerOpt) (TestServer, error) {
 	ts.pgURL.set = make(chan struct{})
 
 	if err := ts.Start(); err != nil {
```
irfansharif, line 279: Is there an error code/string we can match against that would tell us we're running against 19.1 servers (or below) where `start-single-node` is unsupported? Following that, do we have to construct a fresh test server from scratch or can we amend in the cmdArgs into the existing one? Having non trivial code inside an `err` block where err is defined inline (`if err := ...; err != nil { ... }`) is a bit non-standard I think, so I'd be curious to see if we can match against a specific error message and amend things internally.

---

I'm actually not sure how this package gets used, and I suspect what we want in general for cockroach-go is a more "versioned" API in general so that we're able to create a New191TestServer, and so on, but that's too big an ask for someone who knows little about cockroach-go. It's unusual that this code is somewhat version-aware but our APIs don't reflect it. It's either that, or we want individual CRDB releases to pick up a specific tag/branch of this repo. (again, probably too big a yak to shave). I'll defer to you/Yahor for this PR.

rafiss: The exit code is just 1. but we could use the Stderr() output, good call.

Re: making a new testserver, I also wasn't a huge fan, but calling `Start()` alters the internal state of the struct (i.e., `state`, `stdout`, and `stderr` are all changed). So I figured it would be cleaner to initialize it from scratch rather than altering specific fields since that seems a bit brittle.

I definitely agree that this package should just make it easier to pick a version to test with. It's a bigger lift than I have time for now. I made this quicker fix just because currently backports to the release-19.1 branch are broken, and @yuzefovich is trying to get one in.

irfansharif: LGTM (somehow missed this notification).

rafiss:
thank you!

PR #85, opened 2020-07-29
Use `start` if version is <= 19.1

rafiss:
The previous retry logic was broken due to how the package handles
errors.

Now, we check the version up front, and use `start` if necessary.

The version package is copied from the cockroachdb repo.

### DIFF
--- a/testserver/version/version.go
+++ b/testserver/version/version.go
@@ -1,0 +1,239 @@
+// Copyright 2020 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


testserver/version/version.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	if v.metadata != "" {
+		fmt.Fprintf(&b, "+%s", v.metadata)
+	}
+	return b.String()
+}
+
+// versionRE is the regexp that is used to verify that a version string is
+// of the form "vMAJOR.MINOR.PATCH[-PRERELEASE][+METADATA]". This
+// conforms to https://semver.org/spec/v2.0.0.html
```
yuzefovich, line 78: nit: missing period.

rafiss: I'll keep it this way so that the link keeps working.

rafiss:
Going to take the LGTM and merge since this it is a temporary workaround and 19.1 backports are blocked right now.

PR #89, opened 2020-08-11
Read body to reuse TCP connection

yashgandhi-32:
As per docs

```
// HTTP client gurantee that body is always
// no nil. The default HTTP client's Transport may not
// reuse "keep-alive" TCP connections if the Body is
// not read to completion and closed.
```

To ensure http.Client connection reuse

-   Read until complete (ex ioutil.ReadAll(resp.Body)) and close  the body

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -17,6 +17,7 @@ package testserver
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"mime"
 	"net/http"
@@ -90,7 +91,20 @@ func downloadLatestBinary() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer func() { _ = response.Body.Close() }()
+	defer func() {
+
+		// To reuse same tcp connection
+		// Response body is read and closed.
+		// Read the previous response body if it's
+		// small to do connection reuse
+
+		const bodySize = 2 << 20
+
+		if response.ContentLength < bodySize {
+			io.CopyN(ioutil.Discard, response.Body, bodySize)
+		}
+		response.Body.Close()
+	}()
 
 	if response.StatusCode != 200 {
 		return "", fmt.Errorf("error downloading %s: %d (%s)", url, response.StatusCode, response.Status)


testserver/binaries.go
```diff
@@ -90,7 +91,20 @@ func downloadLatestBinary() (string, error) {
...
+
+		// To reuse same tcp connection
+		// Response body is read and closed.
+		// Read the previous response body if it's
+		// small to do connection reuse
+
+		const bodySize = 2 << 20
+
+		if response.ContentLength < bodySize {
+			io.CopyN(ioutil.Discard, response.Body, bodySize)
```
rafiss, line 104: Line 48 is already copying the response.Body. Why is this needed here?

yashgandhi-32: If there's still data pending, the Close() will actually close it and it can't be reused. The default HTTP client's Transport may not reuse HTTP/1.x "keep-alive" TCP connections if the Body is not read to completion and closed. So for small requests up to `2 << 20` bytes data is discarded and it must be ensured to reach EOF

yashgandhi-32: Can also check this for reference 
https://stackoverflow.com/questions/17948827/reusing-http-connections-in-golang

rafiss: Thanks for the reference. One of the comments in that post points out that the Go implementation of `Body.Close()` already takes care of reading the response body fully. See: https://github.com/golang/go/blob/9d23975d89e6cc3df4f2156b2ae0df5d2cef16fb/src/net/http/transfer.go#L979

yashgandhi-32: No Please check below comments where it shared, it does not ensure to close it, it was for post `maxPostHandlerReadBytes` https://github.com/golang/go/blob/9d23975d89e6cc3df4f2156b2ae0df5d2cef16fb/src/net/http/transfer.go#L960

Its mentioned 
"""The code you linked is for the server to read a POST body because a handler isn't required to read or Close the body like a client. So I misspoke and it's not actually a helper for the client at all (IIRC the client will only capture up to the total of the internal buffers, but I haven't followed that code in a long time)"""

rafiss: I see. My next question is about the bodySize. How did you decide on `2 << 20`? That is 2 MB, correct? The files that are being downloaded by this function are much larger than 2 MB, so I don't think the new CopyN code will ever be executed.

And to repeat an earlier question: Line 48 is already copying the response.Body. Why do you need to read it again here?

yashgandhi32: Yes seems to be right but I will test it again, I found some issue earlier not able to find now, closing it.

PR #93, opened 2020-09-09
Add max retry count to retry logic

rafiss:
In the future, this can be made configurable.

### DIFF
--- a/crdb/common.go
+++ b/crdb/common.go
@@ -54,6 +54,9 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 		return err
 	}
 
+	// TODO(rafi): make the maxRetryCount configurable. Maybe pass it in the context?)
+	const maxRetries = 50
+	retryCount := 0
 	for {
 		released := false
 		err = fn()
@@ -75,6 +78,11 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 
 		if retryErr := tx.Exec(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); retryErr != nil {
 			return newTxnRestartError(retryErr, err)
+		}
+
+		retryCount++
+		if retryCount > maxRetries {
+			return newMaxRetriesExceededError(err, maxRetries)
 		}
 	}
 }
--- a/crdb/error.go
+++ b/crdb/error.go
@@ -42,6 +42,21 @@ func newAmbiguousCommitError(err error) *AmbiguousCommitError {
 	return &AmbiguousCommitError{txError{cause: err}}
 }
 
+// MaxRetriesExceededError represents an error caused by retying the transaction
+// too many times, without it ever succeeding.
+type MaxRetriesExceededError struct {
+	txError
+	msg string
+}
+
+func newMaxRetriesExceededError(err error, maxRetries int) *MaxRetriesExceededError {
+	const msgPattern = "retrying txn failed after %d attempts. original error: %s."
+	return &MaxRetriesExceededError{
+		txError: txError{cause: err},
+		msg:     fmt.Sprintf(msgPattern, maxRetries, err),
+	}
+}
+
 // TxnRestartError represents an error when restarting a transaction. `cause` is
 // the error from restarting the txn and `retryCause` is the original error which
 // triggered the restart.


andreimatei:
Why backoff?

rafiss:
it seems like we recommend it here: https://www.cockroachlabs.com/docs/stable/transactions.html 

so I thought the library we suggest to people should have it too

rafiss:
cc @ericharmeling

andreimatei:
That advice is for people not using `savepoint cockroach_restart`. While you're holding locks, the last thing you want to do is backoff.

rafiss:
@andreimatei I see, I did not realize. What are your thoughts on a max retry count?

andreimatei:
Well, on a max retry count more thoughts are less clear. I generally believe in policies for such things (timeouts, limits on the amount of "work" one is willing to put in query) as long as they can be overridden by the client. On the server side we haven't been too successful in adding such policies yet (around retries as well as around range unavailability), but there's going to be a new push in 20.2 (in particular, we want to add control over what happens when a KV operation is waiting on a lock -> the caller should be able to specify a timeout). We also want to continue fighting various starvation scenarios where a transaction doesn't make progress. We've constantly improved such cases - we're now better than we were at leaving locks in place after a failed attempt, we're more sophisticated in changing how we do things when a transaction does retry some operations such that they have higher chances of working the 2nd time around, etc.

So I guess I wouldn't start with putting such a limit in this client library; it's a bit silly to have the limit here for client-directed retries while we don't have the most analogous one on the server side - a limit on the number of auto-retries for a statement (other then the `statement_timeout` session var). But if you want it, I won't stop you.

rafiss:
>  I guess I wouldn't start with putting such a limit in this client library; it's a bit silly to have the limit here for client-directed retries while we don't have the most analogous one on the server side

In fact I would view that as an argument _for_ adding a limit. Without this, there is no way for a user of this library to limit the number of times something gets retried. We've seen multiple support issues where a transaction gets retried thousands of times (or possibly an infinite number of times). I do definitely agree that these scenarios are improving thanks to a lot of hard work, but I also think even if 0.1% of queries can lead to an infinite retry situation, then that's no good. Over time, the infinite retries will build up and starve out other queries, no matter how small that fraction is. That's why there needs to be some global limit somewhere.

I am definitely a fan of things getting better in the next release also, but in the meantime I feel like people need some way of dealing with runaway retry situations. Also, no matter how many improvements are added on the server-side, won't we always need to advise some form of client-side retry logic? Since the client is the one retrying, I think it's reasonable to impose the retry limit in the client code.

Anyway, I mostly wrote all that out just to set some context -- I see that you aren't blocking the addition of a retry limit.

andreimatei (commented):
> Without this, there is no way for a user of this library to limit the number of times something gets retried.

FWIW, there is; the client supplies the closure; they can put whatever limit they want in there (i.e. return an error when they're tired of retrying, or return success without doing anything).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-:-MGtDdes8TcdH4oY5wA1:b-qx3nk6)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andreimatei, @apantel, and @rafiss)

___
*[crdb/common.go, line 86 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MGtEdRTAZpTgjLS1Qdy:-MGtEdRTAZpTgjLS1Qdz:b-lnwk4t) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/2717c4c69ed5f90924cbdebe1f3c56a56004d509/crdb/common.go#L86)):*
> ```Go
>         retryCount++
>         if retryCount > 50 {
>             return err
> ```

wrap this error saying something about the limit
___
*[crdb/common.go, line 89 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MGtEujJ4ZrGSqAlBVuH:-MGtEujJ4ZrGSqAlBVuI:b-uz0pke) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/2717c4c69ed5f90924cbdebe1f3c56a56004d509/crdb/common.go#L89)):*
> ```Go
>         }
>         time.Sleep(sleepTime)
>         if sleepTime < time.Second {
> ```

this goes away, right?

rafiss (commented):
That's fair. I didn't think of that. It still feels cumbersome to me. It means that their closure has to reference some state that they define in their program outside of the closure. To me that feels like violating some sort of encapsulation principle / functional boundary. Also, here's an example of a user complaining that they have to do that hacky sort of thing with one of our other helper libraries which also handles transaction retries: https://github.com/cockroachdb/sqlalchemy-cockroachdb/issues/98

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-:-MGtGF4079IouepchEOM:b-hkqltf)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andreimatei and @apantel)

___
*[crdb/common.go, line 86 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MGtEdRTAZpTgjLS1Qdy:-MGtGCG410-nK5eUHysx:bb6un1j) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/2717c4c69ed5f90924cbdebe1f3c56a56004d509/crdb/common.go#L86)):*

wrap this error saying something about the limit
</blockquote></details>

Done
___
*[crdb/common.go, line 89 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MGtEujJ4ZrGSqAlBVuH:-MGtGCwhCreRyjHpUajO:b-v2wlyu) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/2717c4c69ed5f90924cbdebe1f3c56a56004d509/crdb/common.go#L89)):*

this goes away, right?
</blockquote></details>

Whoops, yes. I never pushed my local changes

rafiss:
This is ready for another review

andreimatei (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-:-MHlLSrgB_ean2iJYIq2:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @andreimatei, @apantel, and @rafiss)

___
*[crdb/error.go, line 53 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MHlLJQJ88UUD8IkGkeM:-MHlLJQJ88UUD8IkGkeN:b-39qtl2) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d51f6bfb0eb263749a30038100b25353d90ad3d1/crdb/error.go#L53)):*
> ```Go
> 
> func newMaxRetriesExceededError(err error, maxRetries int) *MaxRetriesExceededError {
> 	const msgPattern = "retrying txn failed after %d attempts. original error: %s."
> ```

"original" is confusing here, because this is not the error encountered by the first attempt of the transaction, but rather that "last attempt error".

rafiss (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/93)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @andreimatei and @apantel)

___
*[crdb/error.go, line 53 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MHlLJQJ88UUD8IkGkeM:-MI-TfoVE4rXs_Wea2px:bsae0nu) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d51f6bfb0eb263749a30038100b25353d90ad3d1/crdb/error.go#L53)):*

"original" is confusing here, because this is not the error encountered by the first attempt of the transaction, but rather that "last attempt error".
</blockquote></details>

I was using the same phrasing that newTxnRestartError has, but if it becomes an issue we can change both later.

andreimatei (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/93)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @andreimatei, @apantel, and @rafiss)

___
*[crdb/error.go, line 53 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/93#-MHlLJQJ88UUD8IkGkeM:-MI-UELWDDzaV7mdACrS:bh0hbfa) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d51f6bfb0eb263749a30038100b25353d90ad3d1/crdb/error.go#L53)):*

I was using the same phrasing that newTxnRestartError has, but if it becomes an issue we can change both later.
</blockquote></details>

Well but in the `TxnRestartError` case there's no particular reason to believe that there's any other error to talk about then the one causing the `ROLLBACK TO SAVEPOINT` to fail. In the `MaxRetriesExceededError` case, there most certainly is.

PR #96, opened 2020-11-28
Make transaction retry attempts configurable via context value

martingallagher:
This is just a proposal in line with the comment in the code that suggests using the context.

I found the max retries error is quite prevalent when testing on a local dev laptop using mildly concurrent code that puts a lot of contention on a single table.


andreimatei (changes requested):
Passing this option through the context is a particularly obscure way to do it. I think we should start with the higher-level `ExecuteTx` functions (there's one in every sub-package) and add an options param (all these functions already take some opts; I'd change those to be a struct with even more options). 
Besides a retry count limit, consider a timeout option too. I think that's what most people would want. I'm not sure whether the semantics of the timeout should be particularly tied to retries or not.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/96#-:-MNUP0Wn8hZjguCpcQN8:b-iu8rtm)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @martingallagher)

___
*[crdb/common.go, line 29 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/96#-MNUQuPl1U_SxaM1VxQv:-MNUQuPm7gzIIpKYE3wY:bvqb0k6) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/5359cd00f315d4225db6c2936eb15a6295759d75/crdb/common.go#L29)):*
> ```Go
> }
> 
> type maxRetriesKey string
> ```

why a string an not an empty struct? That way you wouldn't be tempted to give it a value below.

martingallagher:
@andreimatei what about variadic options (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) e.g.:

```go
func WithRetries(retries int) Option {
	return func(c *Client) {
		c.retries = retries
	}
}

err := ExecuteTx(
	ctx, conn, txOptions, func(pgx.Tx) error { return nil }, WithRetries(100),
)


err := ExecuteTx(
	ctx, conn, func(pgx.Tx) error { return nil }, WithRetries(100), WithTxOptions(pgx.TxOptions{}),
)
```

A decision would then have to be made what this supports, is it alongside `txOptions pgx.TxOptions` (`WithTxOptions`)? I guess this is a question for the CDB team.

Regarding the timeout, isn't that supported as-is, with `context.WithTimeout`?

andreimatei (commented):
> A decision would then have to be made what this supports, is it alongside txOptions pgx.TxOptions (WithTxOptions)? I guess this is a question for the CDB team.

I'll let @rafiss speak here. I'd unify the existing `txOptions` with the new variadic stuff and bump the module version.

> Regarding the timeout, isn't that supported as-is, with context.WithTimeout?

I think that works when using `pgx` as a driver, but not `lib/pq`. I believe `lib/pq` uses the Postgres cancelation protocol, which CRDB doesn't support. I don't know what the different ORMs use internally.

But even when it does currently "work", it doesn't necessarily do what one might want. It will cancel an outstanding query regardless of whether it's retrying or not. In the context of talking about limiting retries, I was thinking we'd want to put a timeout only on retrying transactions - as a way to bound thrashing. Limiting retries is a pretty shaky thing to begin with, as it doesn't express a very clear user intent (there's different types of retries internal to CRDB that the client doesn't control at the moment, for example), so I'm personally pretty ambivalent towards the whole thing.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/96#-:-MNUhf3H1DJTTuO1Uk0a:b-naox60)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @martingallagher)

rafiss:
thanks for taking this up!

I think variadic functional options is a good approach here.  If you are up for doing a bigger refactor, then I agree that it's time to move the existing `txOptions` in all of the `ExecuteTx` funcs to make use of these functional options as well. (and go to v3 of this library)

I suppose we could just stick on an optional retry count and leave the `txOptions` as-is, and stay backwards-compatible, but that feels a bit too messy to me.

I agree that a timeout option would have additional value and is better than relying on the context timeout. Actually, I think the way pgx implements context cancelation is by closing the connection once the timeout is exceeded, so that "works" but isn't ideal. The timeout option we would want is just: "stop retrying this transaction after X amount of time has passed." No need to do this timeout option in the same PR though.

martingallagher:
I've been toying with this a bit, and I think the TxOptions might need to remain part of the function sigs as they're (subpackage) specific. I have something functionally complete with the txOptions params retained.

It could be worked around by having tx options fields public on an options struct, but this seems against the spirit of functional options.

Any thoughts?

erikgrinaker:
This would be great! We're sometimes seeing pathological behavior where large txns get stuck in infinite retry loops that end up wreaking havoc on the cluster (e.g. cockroachdb/cockroach#62470). While we could implement our own retry logic in the txn callback, that wouldn't give us access to the error that caused the retries, and that info is often useful.

>> Regarding the timeout, isn't that supported as-is, with context.WithTimeout?
>
>I think that works when using pgx as a driver, but not lib/pq. I believe lib/pq uses the Postgres cancelation protocol, which CRDB doesn't support. I don't know what the different ORMs use internally.

Let's use contexts for this. Even though underlying drivers may or may not support this, we should abort as soon as possible on a best-effort basis and add the necessary logic for this (e.g. by closing connections or whatever).

>I've been toying with this a bit, and I think the TxOptions might need to remain part of the function sigs as they're (subpackage) specific. I have something functionally complete with the txOptions params retained.

Variadic options sounds good, I'd just add a subpackage-specific option function for the txOptions as well, something along the lines of:

```go
err := ExecuteTx(
	ctx, conn, func(pgx.Tx) error { return nil }, WithRetries(100), crdbpgx.WithTxOptions(pgx.TxOptions{}),
)
```

Are you willing to pick this back up @martingallagher, or should we just create an issue and have someone else work on it?

rafiss:
@erikgrinaker I'm a bit curious about the infinite retry havoc that you saw. Which version of cockroach-go is used by that test?

In https://github.com/cockroachdb/cockroach-go/pull/93 I added a hard-coded max retry count since I feared exactly the situation you describe. If you import v2.0.8 or v2.1.0 then at least the test might be happier.

Of course, I still advocate for making retries configurable, but just letting you know in case that unblocks you in the short-term.

martingallagher:
@erikgrinaker  @rafiss for what it's worth I've updated my fork / PR with what I've been trialling locally.

erikgrinaker:
Thanks @martingallagher!

>I'm a bit curious about the infinite retry havoc that you saw. Which version of cockroach-go is used by that test?
>
>In #93 I added a hard-coded max retry count since I feared exactly the situation you describe. If you import v2.0.8 or v2.1.0 then at least the test might be happier.

Yeah, I saw that, thanks! CRDB `master` still uses an older version. It wouldn't have helped much in this case, since the txn takes on the order of 10 minutes to execute, so something like 3-5 retries would be appropriate. Ended up just tracking retries manually and using a context time as well (cockroachdb/cockroach#63381), but that won't have access to the retry error.

Issue #98, opened 2021-02-20
Second call to Start() throws error


danieloliveira079:
The current [documentation](https://github.com/cockroachdb/cockroach-go/blob/master/testserver/testserver.go#L30) instructs that the method `Start` should be called after creating a new test server. However, calling this method throws the error: `Start() can only be called once`

https://github.com/cockroachdb/cockroach-go/blob/master/testserver/testserver.go#L30

```
ts, err := testserver.NewTestServer()
if err != nil {
    t.Fatal(err)
}

if err := ts.Start(); err != nil { 
    t.Fatal(err) <== "Start() can only be called once"
}

defer ts.Stop()
```

The first call to the method is happening in https://github.com/cockroachdb/cockroach-go/blob/master/testserver/testserver.go#L328.

Expected:
Remove the following block from he documentation.

```
if err := ts.Start(); err != nil { 
    t.Fatal(err)
}
```

rafiss:
Thank you for noticing this! The docs were not updated to account for https://github.com/cockroachdb/cockroach-go/pull/74

PR #99, opened 2021-02-20
Remove call to Start() from the testserver docs

danieloliveira079:
This PR aims to update the docs in regards to how the test server is used. The is an open issue that reports the details on https://github.com/cockroachdb/cockroach-go/issues/98


rafiss:
Thank you for fixing this!

PR #101, opened 2021-04-08
testserver: run cockroach with Pdeathsig on linux

ajwerner:
If the test crashes for whatever reason and does not run the deferred
cleanup, then we would like very much for the operating system to kill
the subprocess. Linux has nice support for this.

### DIFF
--- a/testserver/exec_linux.go
+++ b/testserver/exec_linux.go
@@ -1,0 +1,30 @@
+// Copyright 2021 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


andreimatei (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/101#-:-MXrUxom2eo9VAfCawyg:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @rafiss)

rafiss:
I don't know why the SIGSEGV is happening but I stumbled upon this: https://github.com/golang/go/issues/27505#issuecomment-713706104

Do you think we need something like that? Do we need `Setpgid=true`?

PR #107, opened 2021-06-04
testserver: fix updated proxy arguments

darinpp:


rafiss (approved):
thank you for this quick fix!

PR #108, opened 2021-08-06
add support for `--store=type=mem` when start a server

ZhouXing19:
This commit enables users to set the database's store path to memory when starting a new test server.


testserver/testserver.go
```diff
@@ -178,6 +179,14 @@ func SecureOpt() TestServerOpt {
 	}
 }
 
+// StoreInMemOpt is a TestServer option that can be passed to NewTestServer
+// to enable storing database in memory
+func StoreInMemOpt() TestServerOpt {
+	return func(args *testServerArgs) {
```
rafiss, line 185: actually, can we make it so storing in memory is the default behavior?

so maybe change this to be a `storeOnDisk` flag which defaults to false and can be made true with an option

testserver/testserver.go
```diff
@@ -292,6 +301,13 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		startCmd = "start"
 	}
 
+	var storeArg string
+	if serverArgs.storeInMem {
+		storeArg = "--store=type=mem"
```
rafiss, line 306: when using a `mem` store, you need to also specify a size: e.g. `--store=type=mem,size=0.2`

we could even consider making it so there's an option to pass in a whole store spec string, so there is more flexibility.

testserver/testserver.go
```diff
@@ -300,7 +316,7 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		"--host=localhost",
 		"--port=0",
 		"--http-port=0",
-		"--store=" + baseDir,
+		storeArg,
```
rafiss, line 319: can you verify the logging behavior with an in-mem store? i think by default the logs go to the console, so we will want to change it so the logs are written to a file instead. i notice that the `logDir` variable is unused right now, so we should probably use it somewhere.


unfortunately the way logging is configured has changed in different versions

in 21.1, i think we use `--log="{file-defaults: {dir: /path/to/log-dir}, sinks: {stderr: {filter: NONE}}}"`
previously, we would use `--log-dir=...`

rafiss (changes requested):

ZhouXing19 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/108)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss)

___
*[testserver/testserver.go, line 185 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/108#-MgRHuRV1lv_ZJdWqqGk-r1-185:-MgqdfPz6Y3Y1mLaOLse:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/32d57677ca885782ff95f419302af1d9e367f673/testserver/testserver.go#L185)):*

actually, can we make it so storing in memory is the default behavior?

so maybe change this to be a `storeOnDisk` flag which defaults to false and can be made true with an option
</blockquote></details>

Done.
___
*[testserver/testserver.go, line 306 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/108#-MgRHuRV1lv_ZJdWqqGk-r1-306:-Mgqe40l1t6TRIywRblK:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/32d57677ca885782ff95f419302af1d9e367f673/testserver/testserver.go#L306)):*

when using a `mem` store, you need to also specify a size: e.g. `--store=type=mem,size=0.2`

we could even consider making it so there's an option to pass in a whole store spec string, so there is more flexibility.
</blockquote></details>

Done.
___
*[testserver/testserver.go, line 319 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/108#-MgRHuRV1lv_ZJdWqqGk-r1-319:-Mgqddz94cpOTOHS62ar:b-4guwh0) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/32d57677ca885782ff95f419302af1d9e367f673/testserver/testserver.go#L319)):*

can you verify the logging behavior with an in-mem store? i think by default the logs go to the console, so we will want to change it so the logs are written to a file instead. i notice that the `logDir` variable is unused right now, so we should probably use it somewhere.


unfortunately the way logging is configured has changed in different versions

in 21.1, i think we use `--log="{file-defaults: {dir: /path/to/log-dir}, sinks: {stderr: {filter: NONE}}}"`
previously, we would use `--log-dir=...`
</blockquote></details>

Made issue #109 to address it.

testserver/testserver.go
```diff
@@ -178,6 +182,27 @@ func SecureOpt() TestServerOpt {
...
+func StoreOnDiskOpt() TestServerOpt {
+	return func(args *testServerArgs) {
+		args.storeOnDisk = true
+	}
+}
+
+// SetStoreMemsize is a TestServer option that can be passed to NewTestServer
+// to set the proportion of available memory that is allocated
+// to the test server.
+func SetStoreMemsize(memSize float64) TestServerOpt {
```
rafiss, line 196: nit: can this be renamed to `SetStoreMemSizeOpt`

(capital S for Size, and end with Opt)

rafiss (approved):
i have a naming request, but after that feel free to merge!

testserver/testserver.go
ZhouXing19: Done.

PR #110, opened 2021-08-13
support downloading the latest stable version of CRDB

### LINKED ISSUE #83: testserver: automatically download only stable cockroachdb versions
Right now testserver downloads any latest builds of Cockroachdb, this includes alphas, betas, etc. 
I think it should download only stable versions by default instead. Or at least add an option for it. 

This was already discussed in https://github.com/cockroachdb/cockroach-go/issues/19. But it seems that there was no resolution and the thread isn't active anymore.

ZhouXing19:
resolves #83

This commit sets the default mode for database downloads
to "Download the latest stable version of CRDB".

We get the latest stable version through
https://register.cockroachdb.com/api/updates, and
save it to a local file with `cockroach-majorv-minorv-patch`
naming pattern in the `tmp` directory. The downloading is
OS-specific.

**Problem unsolved**: the tenant tests with proxy always fail
with stable CRDB, but pass with latest non-stable CRDB.

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -15,9 +15,15 @@
 package testserver
 
 import (
+	"archive/tar"
+	"archive/zip"
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"mime"
 	"net/http"
@@ -28,6 +34,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/gofrs/flock"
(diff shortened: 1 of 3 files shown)


andreimatei (commented):
I feel like there's got to be a better way to know the current version. I think the registration server publishes that information. Have you checked how the DB Console does it when it displays the banner about a new version being available?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-:-MhEdg7wArQ6-E3GwBQg:b-pi4xqf)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss)

ZhouXing19:
@andreimatei Sorry for the late reply. Thanks for the suggestion! I am trying to use [the `updates` endpoint](https://github.com/cockroachlabs/registration#updates) of the CockroachDB cluster to get the latest stable CRDB's binaries. [A related PR](https://github.com/cockroachlabs/registration/pull/271) was made in the Registration repo.

ZhouXing19 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/110)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss)

___
*[testserver/binaries.go, line 119 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk7vWkY1Ypbi3wBax9e:-Mk8KDGW4eFBZwUsNPFu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L119)):*

nit: "saves" instead of "save"
</blockquote></details>

Done.
___
*[testserver/binaries.go, line 120 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk7vYn148oSMpQXzfJE:-Mk8KF7F3gu1P0ZHLP8I:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L120)):*

nit: "returns" instead of "return"
</blockquote></details>

Done.
___
*[testserver/binaries.go, line 122 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk7vc9P5anqhG_HvQV6:-Mk8KG2P1i4KxYHWIoiG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L122)):*

nit: "To download the bleeding edge version of CRDB, set `nonStable` to true."
</blockquote></details>

Done.
___
*[testserver/binaries.go, line 123 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk80ez_9doFf3M88SMJ:-Mk8KHIb1nRe-zmzdJKU:b-nuqfts) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L123)):*

nit: the function should be renamed since `downloadLatestBinary` is not accurate now
</blockquote></details>

Done, changed it to `downloadBinary`.
___
*[testserver/binaries.go, line 136 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk80Pgy1l6L0jec2K8C:-Mk8KVp92rewWEhhxy-J:b467q5n) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L136)):*

should this `exe` logic be part of the `GetDownloadFilename` function?
</blockquote></details>

Done, moved it to `GetDownloadFilename`.
___
*[testserver/binaries.go, line 255 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk81BmxAC1ZpBG7ijtS:-Mk8Kg4vCYreVYtvQI9e:b-12xugi) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L255)):*

nit: would be helpful to name the return values
</blockquote></details>

Done, named it with `filename`.
___
*[testserver/binaries.go, line 305 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-Mk81REN0dURfHzyklNc:-Mk8KbvPDo9vMnB21z9j:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/3ee6f7037e44fef6253daa9de0f3548a5bbdfa20/testserver/binaries.go#L305)):*

this code doesn't handle `zip` right? so i think the comment should just say `tar`
</blockquote></details>

Done.

rafiss (approved):
lgtm!

> Reviewed 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/110#-:-Mkc8sFn5ccOmY2-BY32:b-7nbn20)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss)

PR #116, opened 2021-09-01
Add README for testserver

ZhouXing19:
Add README for testserver


rafiss (approved):
this is very helpful!

PR #119, opened 2021-10-29
rollback on panic

### LINKED ISSUE #76: Prevent transaction from committing on panic.
Currently there is no check for panics within the ExecTx closure. I think it could be useful to have a deferred function rollback the transaction if there is a panic.

ashie1287:
Fixes #76 

~I tried to do the minimum amount of changes to fix the issue.~
UPDATE: did a very small amount of clean up to help me understand a bit better.

There are currently no tests around this. I explored the current tests, and after a while (lots of abstraction), I was able to understand them for the most part. However, it seems that currently (and please correct me if I'm wrong) there are no existing tests actually testing the existing `defer` statement? I would have expected at least some testing of a "non-retryable" error resulting in a rollback, but either there are not any or I couldn't find them.

If someone could advise me on what tests for rollback (whether or not via `panic` or "non-retryable" error, probably should add tests for both) should look like, I'd much appreciate it and am happy to work on them.

Let me know if you have questions, thank you! 🙏

### DIFF
--- a/crdb/common.go
+++ b/crdb/common.go
@@ -39,15 +39,23 @@ type Tx interface {
 // fn is subject to the same restrictions as the fn passed to ExecuteTx.
 func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 	defer func() {
-		if err == nil {
+		r := recover()
+
+		if r == nil && err == nil {
 			// Ignore commit errors. The tx has already been committed by RELEASE.
 			_ = tx.Commit(ctx)
-		} else {
-			// We always need to execute a Rollback() so sql.DB releases the
-			// connection.
-			_ = tx.Rollback(ctx)
+			return
+		}
+
+		// We always need to execute a Rollback() so sql.DB releases the
+		// connection.
+		_ = tx.Rollback(ctx)
+
+		if r != nil {
+			panic(r)
 		}
 	}()
+
 	// Specify that we intend to retry this txn in case of CockroachDB retryable
 	// errors.
 	if err = tx.Exec(ctx, "SAVEPOINT cockroach_restart"); err != nil {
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/cockroachdb/cockroach-go/v2
 go 1.13
 
 require (
-	github.com/gofrs/flock v0.8.1 // indirect
+	github.com/gofrs/flock v0.8.1
```
ashie1287, line 6: Just from `go mod download; go mod tidy` on my machine, happy to revert if I got it wrong.

crdb/common.go
```diff
@@ -58,31 +66,35 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
...
 			// opportunity to react to retryable errors, whereas tx.Commit() doesn't.
-			released = true
 			if err = tx.Exec(ctx, "RELEASE SAVEPOINT cockroach_restart"); err == nil {
 				return nil
 			}
+			releaseFailed = true
 		}
-		// We got an error; let's see if it's a retryable one and, if so, restart.
+
+		// We got an error (i.e. err != nil) if at this point
```
rafiss, line 79: nit: this comment needs to be fixed

rafiss (changes requested):
thanks for this change! it looks good overall -- can you adjust the comment and rebase this?

crdb/common.go
ashie1287: Not 100% sure what you mean, but put it back to exactly what was there. 👍

ashie1287:
@rafiss Thanks for taking a look.

Rebased against `main`, fixed the comment, and actually reverted a few changes, so the change set is now smaller than before (had a misunderstanding of Go named return values).

Just to re-iterate: still no testing around these changes, please advise if you'd like me to add some and if you have any ideas/thoughts on how to accomplish that.

Thanks again! 🙏

rafiss (approved):
this lgtm! i wouldn't worry too much about testing this

rafiss:
If you do want to keep working on the tests for this, here's what you can try:
- implement your own (testing only) version of the `crdb.Tx` interface that can detect when Commit and Rollback have been called.
- call `crdb.ExecuteInTx` with your custom `Tx` and a function that panics. (and another test for a function that returns an error.
- make sure Commit has not been called, and that Rollback has been called.

PR #122, opened 2021-11-15
testserver: fix tenant-client cert arguments

rafiss:
refs https://github.com/cockroachdb/cockroach/issues/71387

Since https://github.com/cockroachdb/cockroach/pull/71248 the server
address must be passed in as a positional argument.

Technically, the version check makes cockroach-go tenant servers not
work for commits between a0c04c620ddb67cf44265eaef61bcd38c13e4fe7 and
d7569dae28d405437f5af670346d45c2d6deb42a, but I think this is an
acceptable breakage since those are unreleased commits.

### DIFF
--- a/testserver/tenant.go
+++ b/testserver/tenant.go
@@ -18,6 +18,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


knz (approved):
cool thanks. There's a typo in the commit/PR title.

PR #123, opened 2022-01-18
Make max retry error show correct message

rafiss:
Previously, it would just show the original cause of the retries, not
the message that says it is giving up on retrying.


otan (approved):
be cool if you could write a quick unit test for this one

rafiss:
added a test, tftr!

PR #125, opened 2022-02-02
fix: user errors.As for error type assertion

aeneasr:
Use proper error unstacking with Go's std libraries. The current code will only check for the first level of nesting whereas Go's std libraries are able to unpack several layers of wrapping.

### DIFF
--- a/crdb/error.go
+++ b/crdb/error.go
@@ -16,23 +16,6 @@ package crdb
 
 import "fmt"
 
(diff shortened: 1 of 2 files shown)


aeneasr:
While I have you, if this change would warrant a release, I'd appreciate it :)

rafiss:
v2.2.7 has been released

aeneasr:
👍

PR #126, opened 2022-02-04
Make it possible to configure maximum retry count

miretskiy:
Make it possible to configure maximum retry count when executing transactions.

### DIFF
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -102,6 +102,26 @@ func Execute(fn func() error) (err error) {
 	}
 }
 
+type txConfigKey struct {}
+
+// WithMaxRetries configures context so that ExecuteTx retries tx specified
+// number of times when encountering retryable errors.
+// Setting retries to 0 will retry indefinitely.
+func WithMaxRetries(ctx context.Context, retries int) context.Context {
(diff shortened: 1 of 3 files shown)


crdb/tx.go
```diff
@@ -101,6 +101,26 @@ func Execute(fn func() error) (err error) {
 	}
 }
 
+type txConfigKey struct {}
+
+// WithMaxRetries configures context so that ExecuteTx retries tx specified
+// number of times when encountering retryable errors.
+// Setting retries to 0 will retry indefinitely.
+func WithMaxRetries(ctx context.Context, retries int) context.Context {
```
rafiss, line 109: could you add a basic test to verify something like

```
ctx = WithMaxRetries(ctx, n)
require.Equal(t, n, numRetriesFromContext(ctx)
```

rafiss (changes requested):

crdb/tx.go
miretskiy: Done.

rafiss (approved):
thanks!

rafiss:
i've tagged this in release v2.2.8

PR #127, opened 2022-02-10
feat: expose testserver BaseDir

aeneasr:
This PR adds `BaseDir() string` which exposes CRDB's root directory which helps us to, for example, load back ups when starting the database.

Please excuse the formatting noise, it is caused by `go fmt -w -s .`

### DIFF
--- a/crdb/tx.go
+++ b/crdb/tx.go
@@ -102,7 +102,7 @@ func Execute(fn func() error) (err error) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


aeneasr:
@rafiss anything else needed to get this merge? The change is straight forward and of great help when we need to access the DB file location

rafiss:
Thanks for your contribution!

PR #130, opened 2022-04-18
Check tx error rather than db error in ExecuteTx

mgovilla:
In ExecuteTx, a transaction (tx) is begun but the error check only looks for errors in the original db object. If the transaction fails to start, the error is not caught immediately.

### DIFF
--- a/crdb/crdbgorm/gorm.go
+++ b/crdb/crdbgorm/gorm.go
@@ -31,8 +31,8 @@ func ExecuteTx(
 	ctx context.Context, db *gorm.DB, opts *sql.TxOptions, fn func(tx *gorm.DB) error,
 ) error {
 	tx := db.WithContext(ctx).Begin(opts)
-	if db.Error != nil {
-		return db.Error
+	if tx.Error != nil {
+		return tx.Error
 	}
 	return crdb.ExecuteInTx(ctx, gormTxAdapter{tx}, func() error { return fn(tx) })
 }
--- a/crdb/crdbgorm/gorm_test.go
+++ b/crdb/crdbgorm/gorm_test.go
@@ -16,6 +16,7 @@ package crdbgorm
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


kernfeld-cockroach:
Could you add a unit test that illustrates an example bug that this would fix?

rafiss (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/130)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mgovilla and @rafiss)

___
*[`crdb/crdbgorm/gorm_test.go` line 18 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/130#-N2J41ub04nTQRuEyess:-N2J41ub04nTQRuEyest:b-n31i4) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/ea5f4b1375316255b4e1c80f476d21e5f635e353/crdb/crdbgorm/gorm_test.go#L18)):*
> ```Go
> 
> import (
> 	"context"
> ```

with my below suggestion, can you import `"errors"` here? (keep this alphabetical)
___
*[`crdb/crdbgorm/gorm_test.go` line 72 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/130#-N2J3-Cn7Mn1KGnjD_Jm:-N2J3-Cn7Mn1KGnjD_Jn:b-l6iznk) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/ea5f4b1375316255b4e1c80f476d21e5f635e353/crdb/crdbgorm/gorm_test.go#L72)):*
> ```Go
> 	err = ExecuteTx(ctx, tx, nil, func(tx *gorm.DB) error {
> 		return gormDB.Exec("SELECT 1+1").Error
> 	})
> ```

could you change this snippet to:
```
	// begin a transaction (bug)
	incorrectTx := gormDB.Begin()

	// attempt to execute a transaction - fails with no valid transaction
	err = ExecuteTx(ctx, incorrectTx, nil, func(tx *gorm.DB) error {
		return tx.Exec("SELECT 1+1").Error
	})
```

i think the naming will make it more clear
___
*[`crdb/crdbgorm/gorm_test.go` line 74 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/130#-N2J3VrRE0rTqjfy7pIt:-N2J3VrS3eB6rJ-BZzJ2:b34xhjh) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/ea5f4b1375316255b4e1c80f476d21e5f635e353/crdb/crdbgorm/gorm_test.go#L74)):*
> ```Go
> 	})
> 
> 	if err == nil {
> ```

can you change this check to:
```
	if !errors.Is(err, gorm.ErrInvalidTransaction) {
```

rafiss:
also, please condense the PR into one commit

rafiss:
Thanks for your contribution!

PR #134, opened 2022-05-08
Define CockroachBinaryPath option for testserver.

gonzojive:
CockroachBinaryPath is a TestServer option that can be passed to NewTestServer
to specify the path of the cockroach binary. This can be used to avoid
downloading cockroach if running tests in an environment with no internet
connection, for instance.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -189,13 +189,24 @@ type TestConfig struct {
 }
 
 type testServerArgs struct {
-	secure       bool
-	rootPW       string  // if nonempty, set as pw for root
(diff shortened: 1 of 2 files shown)


rafiss (changes requested):
thanks for your contribution! could you add a test for this?

gonzojive:
> thanks for your contribution! could you add a test for this?

I added a basic test.

rafiss (changes requested):
thank you! please rebase your changes on top of `origin/master`, and squash your commits into one.

gonzojive:
Done.

rafiss (approved):
thanks!

PR #135, opened 2022-05-14
set cmd.Dir to stop directory pollution

chrisseto:
Previously, invocations of TestServer would results in various _dump
    directories being created by the cockroachdb process. While these
    directories were empty, it was bothersome to see them created and
    show up in file exporers. This commit sets cmd.Dir to a temporary
    directory to avoid polluting the project directory of consumers.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -555,6 +555,11 @@ func (ts *testServerImpl) Start() error {
 		"COCKROACH_TRUST_CLIENT_PROVIDED_SQL_REMOTE_ADDR=true",
 	}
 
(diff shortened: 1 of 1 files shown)


rafiss:
I need to figure out why CI is broken... Trying in https://github.com/cockroachdb/cockroach-go/pull/132

rafiss (changes requested):
nit: could you rebase on master instead of doing a merge?

rafiss (approved):

rafiss (approved):

chrisseto:
TFTR!

PR #136, opened 2022-05-17
cdrb tx: remove libpq specific error

drakkan:
libpq implements SQLState() string method as pgx since 1.10.6.

I think we could also replace the remaining `lib/pq` imports with `pgx` in the future. Fow now is better to keep them and require lib/pq 1.10.6


drakkan:
This patch must be merged after #132 I just applied the changes there to verify that tests pass. I'll now revert that changes using a force-push

rafiss (approved):
hm, will this break anything for users who do not upgrade to the new version of lib/pq?

i wonder if we'd need a new major release if so

drakkan:
> hm, will this break anything for users who do not upgrade to the new version of lib/pq?
> 
> i wonder if we'd need a new major release if so

lib/pq is still imported [here](https://github.com/cockroachdb/cockroach-go/blob/master/crdb/crdbsqlx/sqlx.go#L22) and in the testserver, so it is still in the go.mod. 

I bumped the version in go.mod to 1.10.6, so this should not be a breaking change.
In the future we should replace these imports with pgx and so lib/pq will be removed from go.mod and this will be a breaking change suitable for a major version

rafiss:
i see. thanks for your contribution!

PR #138, opened 2022-06-13
Update crdb binary path to absolute

rimadeodhar:
The teamcity build for cockroachdb repo uses
a relative path for setting a custom cockroach
binary. This change resolves the relative path
to an absolute path. This way updating the command's
base directory does not impact the cockroach binary
look up while running example-orm tests.


rafiss (approved):
is a test needed?

rimadeodhar:
This code is tested by the `TestCockroachBinaryPathOpt`. But I think I will extend it to make the absolute path testing stricter.

PR #140, opened 2022-06-17
Support testserver with multiple nodes.

RichardJCai:
Right now this is only supported in insecure mode.
Start a three node cluster using the ThreeNode TestsServerOpt.


RichardJCai:
Refer to #141

PR #141, opened 2022-06-18
support restarting / upgrading nodes

RichardJCai:


testserver/tenant.go
```diff
@@ -201,27 +202,30 @@ func (ts *testServerImpl) NewTenantServer(proxy bool) (TestServer, error) {
...
 		"--sql-addr=" + sqlAddr,
 		"--http-addr=:0",
-	}
+	}}
 
 	tenant := &testServerImpl{
 		serverArgs: ts.serverArgs,
 		version:    ts.version,
 		state:      stateNew,
+		nodeStates: []int{stateNew},
```
ZhouXing19, line 211: I'm thinking instead of making 4 seperated lists for `args`, `nodeStates`, `cmd` and `listeningURLFile`, can we wrap them into a struct to represent info for each node? 
I.e. can we do
```
testServerImpl{
nodeInfoList: []nodeInfo{node1, node2, node3},
...
}

type nodeInfo struct {
startArgs []string,
nodeState int,
cmd []*exec.Cmd,
listenningURLFile string,
}
```

ZhouXing19 (changes requested):
Some nit comments

testserver/testserver.go
```diff
@@ -469,45 +529,58 @@ func (ts *testServerImpl) BaseDir() string {
...
+	<-ts.pgURL[nodeNum].set
+	return ts.pgURL[nodeNum].u
+}
+
+func (ts *testServerImpl) setPGURLForNode(nodeNum int, u *url.URL) {
+	ts.pgURL[nodeNum].u = u
+	close(ts.pgURL[nodeNum].set)
+}
+
+func (ts *testServerImpl) WaitForNode(nodeNum int) error {
```
ZhouXing19, line 549: nit: what about`WaitForInitFinishForNode`? `WaitForNode` sounds a bit ambiguous for the actual action in this function. Also a comment may help.

ZhouXing19 (changes requested):

testserver/testserver.go
```diff
@@ -287,7 +311,7 @@ var errStoppedInMiddle = errors.New("download stopped in middle")
 // If the download fails, we attempt just call "cockroach", hoping it is
 // found in your path.
 func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
```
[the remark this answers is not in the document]
RichardJCai: It isn't always single node, passing the `ThreeNode TestServerOpt` makes it multinode.

testserver/testserver.go
RichardJCai: Done

testserver/tenant.go
RichardJCai: Done

testserver/testserver.go
```diff
@@ -101,30 +100,41 @@ type TestServer interface {
...
-	cmd              *exec.Cmd
-	cmdArgs          []string
+	mu               sync.RWMutex
+	version          *version.Version
+	serverArgs       testServerArgs
+	state            int
+	nodeStates       []int
+	baseDir          string
+	pgURL            []pgURLChan
+	cmd              []*exec.Cmd
```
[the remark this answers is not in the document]
RichardJCai: Done

ZhouXing19 (approved):
Thanks! Just some very minor nits.

testserver/testserver.go
```diff
@@ -408,30 +443,63 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
...
+		if serverArgs.numNodes > 1 {
+			nodes[i].startCmdArgs = []string{
+				serverArgs.cockroachBinary,
+				startCmd,
+				secureOpt,
+				storeArg + strconv.Itoa(i),
+				fmt.Sprintf("--listen-addr=localhost:%d", 26257+i),
+				fmt.Sprintf("--http-addr=localhost:%d", 8080+i),
+				"--listening-url-file=" + nodes[i].listeningURLFile,
+				fmt.Sprintf("--join=localhost:%d,localhost:%d,localhost:%d", 26257, 26258, 26259),
```
ZhouXing19, line 460: A super minor nit: here for `--join` the params seems hard-coded. I wonder if we want it more general. (Though we may have to involve a tedious for loop)

RichardJCai: I'm going to leave this as hardcoded for now, the ports have to correspond to the `listen-addr` port which always starts at `26257` and ends at `26259` for now so hardcoding always works right now but if we allow the user to pick the `listen-addr` port then we'll have to update this.

PR #142, opened 2022-07-18
tenant: use `-h` to check if tenant scoped client certs available

abarganier:
Previously, whether the test server created tenant-scoped client
certificates for tests was based on a hardcoded version gate. This
was sufficient in the past, but as tenant-scoped client certs are
now being backported to older cockroachdb versions, a more dynamic
approach to determine whether or not these certificates are available
is needed.

This patch adds a mechanism to do so. The new approach runs the
`cockroach cert create-client --help` command to view the available
flags for the current cockroach binary. If the `--tenant-scope` flag
is present in the help text, then we can say with confidence that
tenant scoped client certificates are available. We can use this
to signal the broader system to make use of these certificates when
running tests in secure mode.

This follows the approach used in:
https://github.com/cockroachdb/cockroach/pull/83703

### DIFF
--- a/testserver/tenant.go
+++ b/testserver/tenant.go
@@ -15,6 +15,7 @@
 package testserver
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


rafiss (approved):
nice, this is a great idea

abarganier:
TFTR!

abarganier:
NB: Pushed [tag v2.2.15](https://github.com/cockroachdb/cockroach-go/releases/tag/v2.2.15) with this change.

PR #145, opened 2022-09-06
Require specifying ports when using multiple nodes

RichardJCai:
Previously it was using hardcoded ports, to allow parallel runs we want to be able to specify nodes.

It is on the user to make sure when a node is specified, that port cannot be stolen away during a node restart.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -216,6 +216,7 @@ type testServerArgs struct {
 	storeOnDisk            bool    // to save database in disk
 	storeMemSize           float64 // the proportion of available memory allocated to test server
 	httpPort               int
+	listenAddrPorts        []int
 	testConfig             TestConfig
 	nonStableDB            bool
 	cockroachBinary        string // path to cockroach executable file
@@ -293,6 +294,16 @@ func ExposeConsoleOpt(port int) TestServerOpt {
 	}
 }
 
+// AddPortOpt is a TestServer option that can be passed to NewTestServer to
+// specify the ports for the Cockroach nodes.
+// In the case of restarting nodes, it is up to the user of TestServer to make
+// sure the port used here cannot be re-used.
+func AddPortOpt(port int) TestServerOpt {
+	return func(args *testServerArgs) {
+		args.listenAddrPorts = append(args.listenAddrPorts, port)
+	}
+}
+
 // StopDownloadInMiddleOpt is a TestServer option used only in testing.
 // It is used to test the flock over downloaded CRDB binary.
 // It should not be used in production.
@@ -335,6 +346,19 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		serverArgs.cockroachBinary = *customBinaryFlag
 	} else if customBinaryEnv := os.Getenv("COCKROACH_BINARY"); customBinaryEnv != "" {
 		serverArgs.cockroachBinary = customBinaryEnv
+	}
+
+	// For backwards compatibility, in the 3 node case where no args are
+	// specified, default to ports 26257, 26258, 26259.
+	if serverArgs.numNodes == 3 && len(serverArgs.listenAddrPorts) == 0 {
+		serverArgs.listenAddrPorts = []int{26257, 26258, 26259}
+	} else if serverArgs.numNodes != 1 && len(serverArgs.listenAddrPorts) != serverArgs.numNodes {
+		panic(fmt.Sprintf("need to specify a port for each node using AddPortOpt, got %d nodes, need %d ports",
+			serverArgs.numNodes, len(serverArgs.listenAddrPorts)))
+	}
+
+	if len(serverArgs.listenAddrPorts) == 0 || len(serverArgs.listenAddrPorts) == 1 {
+		serverArgs.listenAddrPorts = []int{0}
 	}
 
 	var err error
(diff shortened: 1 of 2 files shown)


testserver/testserver.go
```diff
@@ -445,19 +463,25 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
...
+			joinArg := fmt.Sprintf("--join=localhost:%s", strings.Join(portArgsStr, ",localhost:"))
 			nodes[i].startCmdArgs = []string{
 				serverArgs.cockroachBinary,
 				startCmd,
 				secureOpt,
 				storeArg + strconv.Itoa(i),
-				fmt.Sprintf("--listen-addr=localhost:%d", 26257+i),
-				fmt.Sprintf("--http-addr=localhost:%d", 8080+i),
+				fmt.Sprintf("--listen-addr=localhost:%d", serverArgs.httpPorts[i]),
+				fmt.Sprintf("--http-addr=localhost:%d", 0),
```
rafiss, line 482: for this to be backwards compatible, i think we want it so still default to 26257/8080 if no port is specified

testserver/testserver.go
```diff
@@ -445,19 +463,25 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
...
 		if serverArgs.numNodes > 1 {
+			joinArg := fmt.Sprintf("--join=localhost:%s", strings.Join(portArgsStr, ",localhost:"))
 			nodes[i].startCmdArgs = []string{
 				serverArgs.cockroachBinary,
 				startCmd,
 				secureOpt,
 				storeArg + strconv.Itoa(i),
-				fmt.Sprintf("--listen-addr=localhost:%d", 26257+i),
-				fmt.Sprintf("--http-addr=localhost:%d", 8080+i),
+				fmt.Sprintf("--listen-addr=localhost:%d", serverArgs.httpPorts[i]),
```
rafiss, line 481: it looks like it's using the HTTP port for the SQL address here

rafiss (changes requested):

testserver/testserver.go
RichardJCai: Oops this should be a different parameter altogether

testserver/testserver.go
RichardJCai: I updated it for 26257 listenaddr but I don't think we actually use the HTTP addr in this case, I think it should be okay to just update it use 0 unless we find that we really need to set it.

RichardJCai: Added a diff parameter to allow people to specify HTTP port as well.

rafiss (changes requested):
this looks fine, just small nits

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/145#-:-NBOINEa6pTYfPjSHGcG:b22bb68)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss, @RichardJCai, and @ZhouXing19)

___
*[`testserver/testserver.go` line 218 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/145#-NBOHCxQDPAoVHo7ektB:-NBOHCxR1HQ5SrQHI7dF:b-ey6m0k) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d1fe11c338ad51a2389999abdf3e2e37cbc642c2/testserver/testserver.go#L218)):*
> ```Go
> 	storeOnDisk            bool    // to save database in disk
> 	storeMemSize           float64 // the proportion of available memory allocated to test server
> 	httpPort               int
> ```

do we still need `httpPort` in addition to `httpPorts`?
___
*[`testserver/testserver.go` line 484 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/145#-NBOHMGF0rSEbOzusIZS:-NBOHMGGDS6vOtaBPd_2:beix1jl) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d1fe11c338ad51a2389999abdf3e2e37cbc642c2/testserver/testserver.go#L484)):*
> ```Go
> 	hostPort := serverArgs.listenAddrPorts[0]
> 	for i, port := range serverArgs.listenAddrPorts {
> 		portArgsStr[i] = fmt.Sprint(port)
> ```

nit: seems like it would be more readable if this was `portArgsStr[i] = fmt.Sprintf("localhost:%d", port)`

then below change to `joinArg := fmt.Sprintf("--join=%s", strings.Join(portArgsStr, ","))`
___
*[`testserver/testserver_test.go` line 474 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/145#-NBOICMG0bfQH4Qz7UP4:-NBOICMHDzlgrKYYKo7Z:bqpwqbw) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d1fe11c338ad51a2389999abdf3e2e37cbc642c2/testserver/testserver_test.go#L474)):*
> ```Go
> 	defer func() {
> 		if err := exec.Command("rm", "-rf", "./temp_binaries").Run(); err != nil {
> 			t.Log(err)
> ```

is it better to keep failing loudly here?

RichardJCai (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/145)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss, @RichardJCai, and @ZhouXing19)

___
*[`testserver/testserver.go` line 218 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/145#-NBOHCxQDPAoVHo7ektB:-NBOjwRM8NnZ159AYK-E:ba7q3zs) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d1fe11c338ad51a2389999abdf3e2e37cbc642c2/testserver/testserver.go#L218)):*

do we still need `httpPort` in addition to `httpPorts`?
</blockquote></details>

Was just being lazy keeping this around for backwards compat, updated it such that `ExposeConsoleOpt` just sets httpPorts[0] to port for backwards compat.
___
*[`testserver/testserver.go` line 484 at r3](https://reviewable.io/reviews/cockroachdb/cockroach-go/145#-NBOHMGF0rSEbOzusIZS:-NBOk7eW921BL2orbX6_:bx0sij5) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/d1fe11c338ad51a2389999abdf3e2e37cbc642c2/testserver/testserver.go#L484)):*

nit: seems like it would be more readable if this was `portArgsStr[i] = fmt.Sprintf("localhost:%d", port)`

then below change to `joinArg := fmt.Sprintf("--join=%s", strings.Join(portArgsStr, ","))`
</blockquote></details>

Yeah sounds good

rafiss (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/145)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss and @ZhouXing19)

PR #146, opened 2022-09-12
Add External IO Dir TestServerOpt

pawalt:
This PR adds `ExternalIODirOpt` in order to let users specify the `--external-io-dir` startup option. This will let users test how their application functions in backup and restore operations without having to use an on-disk testserver.

I also noticed that while `require` was used in `testserver_test.go`, it was not imported in `go.mod`, so I added it in.


RichardJCai (approved):
Nice LGTM

PR #147, opened 2022-09-16
Allow for custom cockroach binary versions

pawalt:
This PR allows users to specify a custom cockroach binary version as opposed to only allowing them to download the latest stable or unstable version.

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -62,7 +62,9 @@ var muslRE = regexp.MustCompile(`(?i)\bmusl\b`)
 // GetDownloadResponse return the http response of a CRDB download.
 // It creates the url for downloading a CRDB binary for current runtime OS,
 // makes a request to this url, and return the response.
-func GetDownloadResponse(nonStable bool) (*http.Response, string, error) {
+// nonStable should only be used if desiredVersion is not specified. If nonStable
+// is used, the latest cockroach binary will be used.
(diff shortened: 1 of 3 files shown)


testserver/binaries.go
```diff
@@ -62,7 +62,7 @@ var muslRE = regexp.MustCompile(`(?i)\bmusl\b`)
 // GetDownloadResponse return the http response of a CRDB download.
 // It creates the url for downloading a CRDB binary for current runtime OS,
```
ZhouXing19, line 63: nit: we may want to update the comment with "`nonStable`is used only if not being given the desired version. In this case, it returns the download response and version for the latest binary."

testserver/binaries.go
```diff
@@ -113,22 +114,22 @@ func GetDownloadResponse(nonStable bool) (*http.Response, string, error) {
...
 			response.StatusCode, response.Status)
 	}
-	return response, latestStableVersion, nil
+	return response, desiredVersion, nil
 }
 
 // downloadBinary saves the latest version of CRDB into a local binary file,
 // and returns the path for this local binary.
 // To download the latest STABLE version of CRDB, set `nonStable` to false.
 // To download the bleeding edge version of CRDB, set `nonStable` to true.
```
ZhouXing19, line 123: nit: update the function comment

ZhouXing19 (approved):
nice job, LGTM! Just some nits

Issue #149, opened 2022-09-28
Add support for jackc/pgx/v5


MoinTom:
`jackc/pgx` introduced a new major version [v5](https://github.com/jackc/pgx/blob/master/CHANGELOG.md#v500)

ralfonso:
This is preventing an update to pgx v5 for my org as well. It would be great to get this into a release.

PR #150, opened 2022-09-28
Add support for`github.com/jackc/pgx/v5`

MoinTom:
Fixes: #149

I assume that it's desired to support `v4` alongside of `v5`, so i introduced a new package `crdbpgx/v5` for the `v5` support. If not desired, i can simply change the original `crdbpgx` to `v5`.


rafiss (approved):
thanks for your contribution!

PR #151, opened 2022-11-01
remove github.com/pkg/errors dependency

### LINKED ISSUE #143: standard package errors should be used instead of archived pkg/errors
Hi,

In https://github.com/cockroachdb/cockroach-go/blob/master/testserver/version/version.go the archived pkg/errors is still being used, it should be replaced with `fmt.Errorf()`.

cornelk:
Closes #143


rafiss (approved):

rafiss:
thanks for your contribution!

PR #153, opened 2022-11-11
Allow single node clusters to override port

pawalt:
Previously, the port option could only be overridden if you were running multiple cockroach nodes. This commit changes the behavior such that you can override from port `0` to any arbitrary port in the single node configuration.


rafiss:
thanks for improving this!

PR #154, opened 2022-11-23
Improve WaitForInit to be more reliable

rafiss:
Using db.Ping() isn't good here, since that can succeed before the database can accept SQL commands.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -652,22 +652,30 @@ func (ts *testServerImpl) setPGURLForNode(nodeNum int, u *url.URL) {
 	close(ts.pgURL[nodeNum].set)
 }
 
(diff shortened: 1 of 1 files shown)


testserver/testserver.go
```diff
@@ -653,15 +653,23 @@ func (ts *testServerImpl) setPGURLForNode(nodeNum int, u *url.URL) {
 }
 
 func (ts *testServerImpl) WaitForInitFinishForNode(nodeNum int) error {
```
ZhouXing19, line 655: nit: `nodeIdx` rather than `nodeNum`?

ZhouXing19 (approved):
LGTM mod a tiny nit

testserver/testserver.go
rafiss: fixed

PR #159, opened 2023-01-31
Support custom host on listen address

pawalt:
Previously, the testserver could only listen on localhost. This meant that there was no way to spin up an instance of the testserver on one machine and access it from another. This commit adds support for custom hosts so that the testserver can be accessed from outside hosts if needed.

For my use case, I need this so that I can have a docker-compose file with two services: one for the testserver and one for my go tests. For the networking to work out, I need to be able to listen on `0.0.0.0` not `localhost`.


ZhouXing19 (approved):
Nice!

pawalt:
TFTR!

PR #167, opened 2023-03-14
Allow non-hardcoded ports for multinode cluster

rafiss:
This required the join and init args to be computed on the fly, since we can no longer infer them until after a node has started.

informs https://github.com/cockroachdb/cockroach/issues/98549
informs https://github.com/cockroachdb/cockroach/issues/98612
informs https://github.com/cockroachdb/cockroach/issues/98594

### DIFF
--- a/testserver/tenant.go
+++ b/testserver/tenant.go
@@ -254,6 +254,7 @@ func (ts *testServerImpl) NewTenantServer(proxy bool) (TestServer, error) {
 	tenantURL := ts.pgURL[0].orig
 	tenantURL.Host = sqlAddr
 	tenant.pgURL = make([]pgURLChan, 1)
+	tenant.pgURL[0].started = make(chan struct{})
 	tenant.pgURL[0].set = make(chan struct{})
 
 	tenant.setPGURL(&tenantURL)
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -123,6 +123,9 @@ type TestServer interface {
 }
 
 type pgURLChan struct {
+	// started will be closed after the start command is executed.
+	started chan struct{}
+	// set will be closed once the URL is available after startup.
 	set chan struct{}
 	u   *url.URL
 	// The original URL is preserved here if we are using a custom password.
@@ -419,17 +422,12 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		serverArgs.cockroachBinary = customBinaryEnv
 	}
 
-	// For backwards compatibility, in the 3 node case where no args are
-	// specified, default to ports 26257, 26258, 26259.
-	if serverArgs.numNodes == 3 && len(serverArgs.listenAddrPorts) == 0 {
-		serverArgs.listenAddrPorts = []int{26257, 26258, 26259}
-	} else if serverArgs.numNodes != 1 && len(serverArgs.listenAddrPorts) != serverArgs.numNodes {
+	if len(serverArgs.listenAddrPorts) == 0 {
+		serverArgs.listenAddrPorts = make([]int, serverArgs.numNodes)
+	}
+	if serverArgs.numNodes != 1 && len(serverArgs.listenAddrPorts) != serverArgs.numNodes {
 		panic(fmt.Sprintf("need to specify a port for each node using AddListenAddrPortOpt, got %d nodes, need %d ports",
 			serverArgs.numNodes, len(serverArgs.listenAddrPorts)))
-	}
-
-	if len(serverArgs.listenAddrPorts) == 0 {
-		serverArgs.listenAddrPorts = []int{0}
 	}
 
 	var err error
@@ -522,13 +520,6 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 	}
 
 	nodes := make([]nodeInfo, serverArgs.numNodes)
-	var initArgs []string
-	joinAddrs := make([]string, 3)
-	hostPort := serverArgs.listenAddrPorts[0]
-	for i, port := range serverArgs.listenAddrPorts {
-		joinAddrs[i] = fmt.Sprintf("localhost:%d", port)
-	}
-
 	if len(serverArgs.httpPorts) == 0 {
 		serverArgs.httpPorts = make([]int, serverArgs.numNodes)
 	}
@@ -551,7 +542,6 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		nodes[i].listeningURLFile = filepath.Join(nodeBaseDir, "listen-url")
 		nodes[i].state = stateNew
 		if serverArgs.numNodes > 1 {
-			joinArg := fmt.Sprintf("--join=%s", strings.Join(joinAddrs, ","))
 			nodes[i].startCmdArgs = []string{
 				serverArgs.cockroachBinary,
 				startCmd,
@@ -568,7 +558,6 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 					serverArgs.httpPorts[i],
 				),
 				"--listening-url-file=" + nodes[i].listeningURLFile,
-				joinArg,
 				"--external-io-dir=" + serverArgs.externalIODir,
 			}
 		} else {
@@ -589,11 +578,10 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 
 	// We only need initArgs if we're creating a testserver
 	// with multiple nodes.
-	initArgs = []string{
+	initArgs := []string{
 		serverArgs.cockroachBinary,
 		"init",
 		secureOpt,
-		fmt.Sprintf("--host=localhost:%d", hostPort),
 	}
 
 	states := make([]int, serverArgs.numNodes)
@@ -611,6 +599,10 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		nodes:       nodes,
 	}
 	ts.pgURL = make([]pgURLChan, serverArgs.numNodes)
+	for i := range ts.pgURL {
+		ts.pgURL[i].started = make(chan struct{})
+		ts.pgURL[i].set = make(chan struct{})
+	}
 
 	if err := ts.Start(); err != nil {
 		return nil, fmt.Errorf("%s Start failed: %w", testserverMessagePrefix, err)
@@ -676,9 +668,10 @@ func (ts *testServerImpl) setPGURLForNode(nodeNum int, u *url.URL) {
 }
 
 func (ts *testServerImpl) WaitForInitFinishForNode(nodeIdx int) error {
+	pgURL := ts.PGURLForNode(nodeIdx).String()
 	for i := 0; i < ts.serverArgs.initTimeoutSeconds*10; i++ {
 		err := func() error {
-			db, err := sql.Open("postgres", ts.PGURLForNode(nodeIdx).String())
+			db, err := sql.Open("postgres", pgURL)
 			if err != nil {
 				return err
 			}
@@ -695,7 +688,7 @@ func (ts *testServerImpl) WaitForInitFinishForNode(nodeIdx int) error {
 		if err == nil {
 			return nil
 		}
-		log.Printf("%s: WaitForInitFinishForNode %d: Trying again after error: %v", testserverMessagePrefix, nodeIdx, err)
+		log.Printf("%s: WaitForInitFinishForNode %d (%s): Trying again after error: %v", testserverMessagePrefix, nodeIdx, pgURL, err)
 		time.Sleep(time.Millisecond * 100)
 	}
 	log.Printf(
@@ -867,7 +860,10 @@ func (ts *testServerImpl) Stop() {
 }
 
 func (ts *testServerImpl) CockroachInit() error {
-	ts.initCmd = exec.Command(ts.initCmdArgs[0], ts.initCmdArgs[1:]...)
+	// The port must be computed here, since it may not be known until after
+	// a node is started (if the listen port is 0).
+	args := append(ts.initCmdArgs, fmt.Sprintf("--host=localhost:%s", ts.PGURL().Port()))
+	ts.initCmd = exec.Command(args[0], args[1:]...)
 	ts.initCmd.Env = []string{
 		"COCKROACH_MAX_OFFSET=1ns",
 		"COCKROACH_TRUST_CLIENT_PROVIDED_SQL_REMOTE_ADDR=true",
@@ -880,7 +876,7 @@ func (ts *testServerImpl) CockroachInit() error {
 
 	err := ts.initCmd.Start()
 	if ts.initCmd.Process != nil {
-		log.Printf("process %d started: %s", ts.initCmd.Process.Pid, strings.Join(ts.initCmdArgs, " "))
+		log.Printf("process %d started: %s", ts.initCmd.Process.Pid, strings.Join(args, " "))
 	}
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


ZhouXing19 (approved):
LGTM! I just left comments about adding more logs

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-:-NQaT0pv4dPk6fh9TVNG:b-cz1ye2)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @rafiss, @renatolabs, and @smg260)

___
*[`testserver/testservernode.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaR34i6tLJF3hWu0nB:-NQaR34i6tLJF3hWu0nC:b-c8t55c) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L74)):*
> ```Go
> 			joinAddrs = append(joinAddrs, fmt.Sprintf("localhost:%s", ts.pgURL[otherNodeID].u.Port()))
> 		case <-time.After(5 * time.Second):
> 			// If the other node hasn't started yet, don't add the join arg.
> ```

nit: might be helpful for future debugging if we can log something here

And should we also retry connecting to this node?
___
*[`testserver/testservernode.go` line 78 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaSLhN9ByrwiN5U9F1:-NQaSLhN9ByrwiN5U9F2:blxh865) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L78)):*
> ```Go
> 	}
> 	joinArg := fmt.Sprintf("--join=%s", strings.Join(joinAddrs, ","))
> ```

nit: can we add a log after exiting the loop, just to show eventually what nodes are going to be joined? 
or maybe more generally just print out the whole `startCmd` right before executing it. 
Just for the debugging purpose. 
___
*[`testserver/testservernode.go` line 84 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaRf8K3acooPFOxAod:-NQaRf8K3acooPFOxAoe:barwfrb) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L84)):*
> ```Go
> 			// The start command always requires a --join arg, so we fake one
> 			// if we don't have any yet.
> 			joinArg = "--join=localhost:0"
> ```

just curious why we need this ...? Can't we just don't set the join flag?

ZhouXing19:
*[`testserver/testservernode.go` line 127 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaTJmL0gicd0s7wvB1:-NQaTJmL0gicd0s7wvB2:bzbtp16) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L127)):*
> ```Go
> 	err := currCmd.Start()
> 	if currCmd.Process != nil {
> 		log.Printf("process %d started: %s", currCmd.Process.Pid, strings.Join(args, " "))
> ```

Should we also include the env vars in the log?

ZhouXing19:
*[`testserver/testservernode.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaR34i6tLJF3hWu0nB:-NQaUdHw4Bs6kj0KPL7u:b-otxqvz) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L74)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

nit: might be helpful for future debugging if we can log something here

And should we also retry connecting to this node?
</blockquote></details>

nvm I just saw the log in the `if currCmd.Process != nil {}` below

ZhouXing19:
*[`testserver/testservernode.go` line 78 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaSLhN9ByrwiN5U9F1:-NQaUkFPAbvINLREneiD:bqmw1lg) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L78)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

nit: can we add a log after exiting the loop, just to show eventually what nodes are going to be joined? 
or maybe more generally just print out the whole `startCmd` right before executing it. 
Just for the debugging purpose. 
</blockquote></details>

ditto, just saw the log below

renatolabs (commented):
Thank you for these fixes! Mostly LGTM except for a few questions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-:-NQae8nU-ADK1nre9OVz:b-34yeco)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @rafiss, and @smg260)

___
*[`testserver/testservernode.go` line 71 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQadhhe5kmU6e45eCAF:-NQadhhgCIt_prCAeHSn:b-m3gdsf) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L71)):*
> ```Go
> 		}
> 		select {
> 		case <-ts.pgURL[otherNodeID].set:
> ```

What are the concurrency guarantees we attempt to make with `testserver`? I see there's some effort in these files to guard against concurrent access, but with this code, it would be problematic if we called `StartNode` and `StopNode` concurrently. Is that a problem?
___
*[`testserver/testservernode.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQacmON2KeFaveu-jMT:-NQacmON2KeFaveu-jMU:b-4s8mtj) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L73)):*
> ```Go
> 		case <-ts.pgURL[otherNodeID].set:
> 			joinAddrs = append(joinAddrs, fmt.Sprintf("localhost:%s", ts.pgURL[otherNodeID].u.Port()))
> 		case <-time.After(5 * time.Second):
> ```

IIUC, this "timeout" is here to filter out nodes that have not been started yet, right? Is t here any way we could reflect that with `ts.pgURL` instead? It seems right now we are creating a fixed-size slice on startup, meaning there's no way, at this point, to know which nodes we have attempted to start, and which we have not. IMO, this would be more reliable if we knew what nodes we know we should wait for, and then fail if we don't hear their PGURL after some time (5s or 10s).

Another thing is that  if we wanted to start a (say) 10-node server, wouldn't it mean that we would spend a lot of time sleeping in this loop?

rafiss (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @renatolabs, @smg260, and @ZhouXing19)

___
*[`testserver/testservernode.go` line 71 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQadhhe5kmU6e45eCAF:-NQb0ksjAbzBFL90hN-J:b2py7z) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L71)):*
<details><summary><i>Previously, renatolabs (Renato Costa) wrote…</i></summary><blockquote>

What are the concurrency guarantees we attempt to make with `testserver`? I see there's some effort in these files to guard against concurrent access, but with this code, it would be problematic if we called `StartNode` and `StopNode` concurrently. Is that a problem?
</blockquote></details>

there's one test called `TestRestartNodeParallel`, and i took that to be the "contract" offered by this package. 

it does concurrently call StopNode and StartNode, but the _same_ node is never started and stopped concurrently.
___
*[`testserver/testservernode.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQacmON2KeFaveu-jMT:-NQb0NyCAbmykB7Sh-Bm:b-jaoq4h) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L73)):*
> IIUC, this "timeout" is here to filter out nodes that have not been started yet, right?

that's right.

> Is there any way we could reflect that with ts.pgURL instead? 

hm, that makes sense. we can add another channel there

> Another thing is that if we wanted to start a (say) 10-node server, wouldn't it mean that we would spend a lot of time sleeping in this loop?

yeah i did notice this. initially i had this as `case default:` but that made the `TestRestartNodeParallel` test (see above) fail, since everything would be unable to find a join port.
___
*[`testserver/testservernode.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaR34i6tLJF3hWu0nB:-NQb1ZywCYwMgEWvXhhV:bb7f2mf) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L74)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

nvm I just saw the log in the `if currCmd.Process != nil {}` below
</blockquote></details>

done
___
*[`testserver/testservernode.go` line 78 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaSLhN9ByrwiN5U9F1:-NQb1_h65IpzcjzvQ8dH:bb7f2mf) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L78)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

ditto, just saw the log below
</blockquote></details>

done
___
*[`testserver/testservernode.go` line 84 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaRf8K3acooPFOxAod:-NQb0GZx7nYrw_lJRt1z:bwjvycd) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L84)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

just curious why we need this ...? Can't we just don't set the join flag?
</blockquote></details>

The `start` command required the join flag
___
*[`testserver/testservernode.go` line 127 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQaTJmL0gicd0s7wvB1:-NQb1pNb1to-3WxEbic5:b-sf8yw5) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/c087468e66494d813f304e9d455d3cc29975ba06/testserver/testservernode.go#L127)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

Should we also include the env vars in the log?  
</blockquote></details>

sure, sgtm

renatolabs (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @rafiss, @smg260, and @ZhouXing19)

___
*[`testserver/testserver.go` line 127 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQb9yH46X8HmcLoA5ep:-NQb9yH46X8HmcLoA5eq:b-sshfbu) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/94807f6139a9b1c624e989912917aedf02f360af/testserver/testserver.go#L127)):*
> ```Go
> type pgURLChan struct {
> 	// started will be closed after the start command is executed.
> 	started chan struct{}
> ```

Wouldn't this be more clearly expressed with a `bool`? What am I missing?

rafiss (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @renatolabs, @smg260, and @ZhouXing19)

___
*[`testserver/testserver.go` line 127 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQb9yH46X8HmcLoA5ep:-NQbB0Sq99Wu9vn_y7kW:bbqvpng) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/94807f6139a9b1c624e989912917aedf02f360af/testserver/testserver.go#L127)):*
<details><summary><i>Previously, renatolabs (Renato Costa) wrote…</i></summary><blockquote>

Wouldn't this be more clearly expressed with a `bool`? What am I missing?
</blockquote></details>

it's possible for two different goroutines to access this. e.g. in `TestRestartNodeParallel` one goroutine could be writing at the same time as reading the bool. or do you mean use an `AtomicBool`?

renatolabs (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @rafiss, @smg260, and @ZhouXing19)

___
*[`testserver/testserver.go` line 127 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQb9yH46X8HmcLoA5ep:-NQbCVX_4bbSbOQ_lUrR:b-m96ns0) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/94807f6139a9b1c624e989912917aedf02f360af/testserver/testserver.go#L127)):*

it's possible for two different goroutines to access this. e.g. in `TestRestartNodeParallel` one goroutine could be writing at the same time as reading the bool. or do you mean use an `AtomicBool`? 
</blockquote></details>

I meant a regular `bool`. The testserver keeps one `pgURLChan` per node, and it seems our contract is that we don't call `StartNode` and `StopNode` for the same node concurrently. When do we have a race?

rafiss (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/167)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @pawalt, @renatolabs, @smg260, and @ZhouXing19)

___
*[`testserver/testserver.go` line 127 at r2](https://reviewable.io/reviews/cockroachdb/cockroach-go/167#-NQb9yH46X8HmcLoA5ep:-NQerIT15ymNUKc8F50B:b3c8aj8) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/94807f6139a9b1c624e989912917aedf02f360af/testserver/testserver.go#L127)):*
<details><summary><i>Previously, renatolabs (Renato Costa) wrote…</i></summary><blockquote>

I meant a regular `bool`. The testserver keeps one `pgURLChan` per node, and it seems our contract is that we don't call `StartNode` and `StopNode` for the same node concurrently. When do we have a race?
</blockquote></details>

We have a race because calling `StartNode` on one node makes this code look at `pgURLChan` from another node.

Suppose node 1 is started concurrently with node 2 being stopped then:
- node 1 could be reading the value here: https://github.com/rafiss/cockroach-go/blob/94807f6139a9b1c624e989912917aedf02f360af/testserver/testservernode.go#L73
- node 2 could be writing the value here: https://github.com/rafiss/cockroach-go/blob/94807f6139a9b1c624e989912917aedf02f360af/testserver/testservernode.go#L44

PR #168, opened 2023-04-13
testserver: add support for the --locality flag

JeffSwenson:
This commit adds the `testserver.LocalityFlagsOpt` option. The flag makes it possible to test multi-region primitives with the test server.


chrisseto (approved):
LGTM!

PR #170, opened 2023-05-26
testserver: introduce CockroachLogsDirOpt

renatolabs:
This new parameter allows callers to specify a root directory where logs for cockroach processes created by testserver will be located.

Previously, when calling `testserver.Stop()`, *all* resources would be deleted, including cockroach logs. By allowing the caller to specify a custom directory, we allow the possibility for the logs to outlive the cockroach processes, which is especially useful when debugging test failures.


rafiss (approved):
this is a great addition, thanks!

PR #172, opened 2023-10-11
testserver: expose `NoFileCleanup` option to allow retention of files after `Stop`

janmejay:
Retaining files after `Stop` can help debug some test-failures because
 one can start the server back up (off the store-dir left behind after
 failure). Looking at the data the failing test left behind can help
 hypothesize better.

### DIFF
--- a/testserver/testserver_test.go
+++ b/testserver/testserver_test.go
@@ -234,6 +234,12 @@ func TestRunServer(t *testing.T) {
 				)
 			},
 		},
+		{
+			name: "No File Cleanup",
+			instantiation: func(t *testing.T) (*sql.DB, func()) {
+				return testserver.NewDBForTest(t, testserver.NoFileCleanup())
+			},
+		},
 	} {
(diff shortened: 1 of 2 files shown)


testserver/testserver_test.go
```diff
@@ -234,6 +234,12 @@ func TestRunServer(t *testing.T) {
 				)
 			},
 		},
+		{
+			name: "No File Cleanup",
+			instantiation: func(t *testing.T) (*sql.DB, func()) {
```
rafiss, line 239: is there any way for this test to assert that the directory still exists at the end?

rafiss (changes requested):
thanks for your contribution!

testserver/testserver_test.go
janmejay: Added a dedicated test `TestBaseDirIsPreservedWhenNoFileCleanupRequested` (but also had to add some additional machinery to power it).

janmejay:
@rafiss does this look alright?

PR #173, opened 2023-11-17
testserver: short circuit downloading before making a request

pawalt:
Previously, the testserver would always make a get request to download
the cockroach binary. This commit short circuits the download logic to
avoid making the initial request if the file already exists.

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -57,12 +57,12 @@ const updatesUrl = "https://register.cockroachdb.com/api/updates"
 
 var muslRE = regexp.MustCompile(`(?i)\bmusl\b`)
 
-// GetDownloadResponse return the http response of a CRDB download.
-// It creates the url for downloading a CRDB binary for current runtime OS,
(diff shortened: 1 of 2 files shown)


rickystewart (approved):

testserver/binaries.go
```diff
@@ -124,19 +134,30 @@ func GetDownloadResponse(desiredVersion string, nonStable bool) (*http.Response,
...
 	if err != nil {
 		return "", err
 	}
 
-	defer func() { _ = response.Body.Close() }()
+	filename, err := GetDownloadFilename(desiredVersion)
+	if err != nil {
+		return "", err
+	}
+	localFile := filepath.Join(os.TempDir(), filename)
```
rafiss, line 146: curious how the short-circuiting will work. won't localFile be different each time this is called, since os.TempDir would return a different directory?

rickystewart: No. You may be thinking of `MkdirTemp`.

Issue #174, opened 2024-01-19
package version logic points to unavailable cockroachdb version


elv-todd:
## Summary

This package's "latest" version logic cannot properly compute CockroachDB latest binaries.

There are different dates for "Dedicated from Self-hosted binaries" vs "Self-hosted binaries"; this package needs the latter.

## Problem description

this package is broken as of yesterday, 17 Jan 2024.  At that time, cockroachdb released `v23.2.0`, but, the binary is only "Available in CockroachDB Dedicated."  This package needs self-hosted binaries, but, "Self-hosted binaries available February 5."

the error logs you will see will be:
```
Step #3: 2024/01/19 00:42:09 cockroach-go testserver: Failed to fetch latest binary: error downloading https://binaries.cockroachdb.com/cockroach-v23.2.0.linux-amd64.tgz: 404 (404 Not Found) attempting to use cockroach 
```

and this is caused by
```
const updatesUrl = "https://register.cockroachdb.com/api/updates"
```
returning
```
{
  "version": "23.2.0",
  "details": "https://www.cockroachlabs.com/docs/releases/v23.2#v23-2-0"
}
```

## Proposed Solution

use `https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz` not an explicit version

is currently `cockroach-v23.1.13.linux-amd64`

## Details

https://www.cockroachlabs.com/docs/releases/ shows

`v23.2.0 Latest 2024-01-17	Available in CockroachDB Dedicated. Self-hosted binaries available February 5.`

looks like we need v23.1.13; the version either needs to be configurable or the selection logic needs to be adapted. This is what we want to see instead:

```
Step #3: 2024/01/17 20:37:39 GET https://binaries.cockroachdb.com/cockroach-v23.1.13.linux-amd64.tgz
```

rafiss:
Thanks for this report! I'll make a fix for this.

elv-todd:
thanks!  nice work

PR #175, opened 2024-01-19
Switch to docs page to determine latest stable version

rafiss:
fixes https://github.com/cockroachdb/cockroach-go/issues/174

The code was previously looking at the /api/updates endpoint on the register.cockroachdb.com server. However, that endpoint returns Cloud Only releases, which cannot be downloaded from binaries.cockroachdb.com.

Now we use the releases list maintained by the docs team, which lets us filter out undesired releases.

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -19,7 +19,6 @@ import (
 	"archive/zip"
 	"bytes"
 	"compress/gzip"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


rail (approved):

elv-todd:
thanks!  to point to it do we use 

```
go get github.com/cockroachdb/cockroach-go/v2@master
```

or are we waiting on a release?

rafiss:
v2.3.6 of this library has this fix

PR #180, opened 2025-02-05
crdb: Add ExecuteCtx for configurable retry limits

sean-:
Adds ExecuteCtx which allows configuring maximum retries through context, alongside the existing Execute function. This enables better control over retry behavior while maintaining compatibility with code using the original Execute.

The new function helps prevent infinite retry loops in environments where bounded retry counts are required, while preserving the existing error handling and retry semantics.


sean-:
Also, FYI re: the reason for the toolchain bumping and limiting the Go support matrix is because various tools were segfaulting on older versions of Go.

rail (approved):
LGTM

rafiss (approved):
> Reviewed 2 of 3 files at r4, 1 of 1 files at r9, 2 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/180)* status: <span>:shipit:</span> complete! 0 of 0 LGTMs obtained (waiting on @sean-)

PR #181, opened 2025-05-15
testserver: fix URL file polling code

RaduBerinde:
The asynchronous code that polls for the URL file leads to very opaque failures. This change reworks this code to waiting synchronously, so that we know the server is started before we return. In cases of failure, we now get more useful errors.

### DIFF
--- a/testserver/testserver.go
+++ b/testserver/testserver.go
@@ -760,34 +760,44 @@ func (ts *testServerImpl) WaitForInit() error {
 }
 
 func (ts *testServerImpl) pollListeningURLFile(nodeNum int) error {
+	log.Printf("node %d: waiting for listening URL file %q\n", nodeNum, ts.nodes[nodeNum].listeningURLFile)
 	var data []byte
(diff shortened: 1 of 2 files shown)


testserver/testserver.go
```diff
@@ -760,28 +760,33 @@ func (ts *testServerImpl) WaitForInit() error {
...
-		if len(data) == 0 {
-			time.Sleep(100 * time.Millisecond)
-			continue
-		} else if err == nil {
+		if err == nil && len(data) > 0 {
 			break
-		} else if !os.IsNotExist(err) {
-			return fmt.Errorf("unexpected error while reading listening URL file: %w", err)
 		}
+		if err != nil && !os.IsNotExist(err) {
```
srosenberg, line 777: Nit: left conjunct is a tautology, it can be dropped.

srosenberg (approved):

testserver/testserver.go
RaduBerinde: It's not, when data is empty. I'll refactor a bit so it's less subtle.

TFTR!

srosenberg: Right, 🤦 .

PR #182, opened 2025-06-09
Flexible retry backoff policies

MattWhelan:
Extends the existing retry limit mechanism to allow for delays between retries. Includes implementations for fixed delays and exponential backoff.


crdb/retry.go
```diff
@@ -0,0 +1,110 @@
...
+			// There's no max delay. Giving up is probably better in
+			// practice than using a 290-year MAX_INT delay.
+			return 0, newMaxRetriesExceededError(err, tryCount)
+		}
+		return delay, nil
+	}
+}
+
+// Vargo converts a go-retry style Delay provider into a RetryPolicy
+func Vargo(fn func() VargoBackoff) RetryPolicy {
```
rafiss, line 82: I'm wondering if we could think of a different name to go with here. Users of cockroach-go who don't already know about the sethvargo/go-retry library may be confused to see this.

crdb/common.go
```diff
@@ -87,14 +87,29 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
...
-		if rollbackErr := tx.Exec(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); rollbackErr != nil {
-			return newTxnRestartError(rollbackErr, err)
+		// We have a retryable error. Check the retry policy.
+		delay, retryErr := retryFunc(err)
+		if delay > 0 && retryErr == nil {
+			// We don't want to hold locks while waiting for a backoff, so restart the entire transaction
+			if restartErr := tx.Exec(ctx, "ROLLBACK"); restartErr != nil {
+				return newTxnRestartError(restartErr, err)
+			}
+			if restartErr := tx.Exec(ctx, "BEGIN"); restartErr != nil {
```
rafiss, line 97: This seems like it could have issues. Some libraries may not like it if we use the transaction after rolling it back. For example, I see this in the pgx code where it returns an error if you try to use a transaction that was already rolled back: https://github.com/jackc/pgx/blob/ecc9203ef42fbba50507e773901b5aead75288ef/tx.go#L205-L224

crdb/tx.go
```diff
@@ -153,29 +158,49 @@ func ExecuteCtx(ctx context.Context, fn ExecuteCtxFunc, args ...interface{}) (er
...
-
-	return newMaxRetriesExceededError(err, maxRetries)
 }
 
 type txConfigKey struct{}
 
 // WithMaxRetries configures context so that ExecuteTx retries tx specified
 // number of times when encountering retryable errors.
-// Setting retries to 0 will retry indefinitely.
+// Setting retries to 0 will not retry: the transaction will be tried only once.
```
rafiss: I think we should try to avoid breaking the API here. This changes the behavior of `WithMaxRetries(ctx, 0)` from "retry indefinitely" to "no retries," which would affect any existing code relying on this.

Maybe we could add a separate `WithNoRetries()` function, and change something about the internal representation so that the outward facing API remains the same.

crdb/common.go
```diff
@@ -82,13 +87,35 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
...
```
rafiss, line 110: This still doesn't seem valid to me -- if we reuse the same `tx` object here, it will not succeed when sending BEGIN right after a ROLLBACK. (see https://github.com/jackc/pgx/blob/ecc9203ef42fbba50507e773901b5aead75288ef/tx.go#L205-L224)

crdb/common.go
```diff
@@ -82,13 +87,35 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 			return err
 		}
 
-		if rollbackErr := tx.Exec(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); rollbackErr != nil {
-			return newTxnRestartError(rollbackErr, err)
+		// We have a retryable error. Check the retry policy.
+		delay, retryErr := retryFunc(err)
+		if delay > 0 && retryErr == nil {
```
rafiss, line 96: I don't follow this. Why are we doing a retry if `retryErr` is nil?

crdb/common.go
```diff
@@ -82,13 +87,35 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
...
+			// We don't want to hold locks while waiting for a backoff, so restart the entire transaction
+			if restartErr := tx.Exec(ctx, "ROLLBACK"); restartErr != nil {
+				return newTxnRestartError(restartErr, err)
```
rafiss: This is the wrong error to return here -- as per this line, when this error is displayed, it will say that there was an issue with ROLLBACK TO SAVEPOINT, which is not what this line is doing: https://github.com/cockroachdb/cockroach-go/blob/c7879874023d8e8cf21d7949df1510fb967a8f0a/crdb/error.go#L70

(This is also the wrong error to return on lines 98 and 101.)

crdb/retry_test.go
```diff
@@ -0,0 +1,72 @@
...
+		t.Errorf("wrong number of retries: expected %d, got %d", len(expectedDelays), len(actualDelays))
+	}
+	for i, delay := range actualDelays {
+		expected := expectedDelays[i]
+		if delay != expected {
+			t.Errorf("wrong delay at index %d: expected %d, got %d", i, expected, delay)
+		}
+	}
+}
+
```
rafiss, line 69: I think we should also have tests for edge cases like zero BaseDelay, negative RetryLimit, and MaxDelay < BaseDelay.

crdb/common.go
```diff
@@ -82,13 +87,35 @@ func ExecuteInTx(ctx context.Context, tx Tx, fn func() error) (err error) {
 			return err
 		}
 
-		if rollbackErr := tx.Exec(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); rollbackErr != nil {
-			return newTxnRestartError(rollbackErr, err)
+		// We have a retryable error. Check the retry policy.
+		delay, retryErr := retryFunc(err)
```
rafiss, line 91: before we enter the retry logic, we should check for context cancellation:

```
if err := ctx.Err(); err != nil {
  return err
}
```

crdb/retry.go
```diff
@@ -0,0 +1,110 @@
...
+type RetryFunc func(err error) (time.Duration, error)
+
+// RetryPolicy constructs a new instance of a RetryFunc for each transaction
+// it is used with. Instances of RetryPolicy can likely be immutable and
+// should be safe for concurrent calls to NewRetry.
+type RetryPolicy interface {
+	NewRetry() RetryFunc
+}
+
+type LimitBackoffRetryPolicy struct {
```
rafiss, line 108: This is public-facing so should have a comment explaining what it is and how to use it.

crdb/retry.go
```diff
@@ -0,0 +1,110 @@
...
+		tryCount++
+		if tryCount > l.RetryLimit {
+			return 0, newMaxRetriesExceededError(err, l.RetryLimit)
+		}
+		return l.Delay, nil
+	}
+}
+
+// ExpBackoffRetryPolicy implements RetryPolicy using an exponential backoff with optional
+// saturation.
```
rafiss: I don't follow what "saturation" means here. Can we write this comment to explain more how to use this?

crdb/retry_test.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"testing"
+	"time"
+)
+
+func assertDelays(t *testing.T, policy RetryPolicy, expectedDelays []time.Duration) {
+	actualDelays := make([]time.Duration, 0, len(expectedDelays))
+	rf := policy.NewRetry()
+	for {
+		delay, err := rf(nil)
```
rafiss, line 30: We're always passing a `nil` error here -- can we add tests that check the behavior with a non-nil error?

crdb/common.go
sravotto: Done.

crdb/common.go
sravotto: I added documentation on the RetryFunc function definition to clarify the semantics.  Implementation of RetryFunc may return errors (typically MaxRetriesExceededError) to stop the retry process.

crdb/common.go
sravotto: Done.  Passing the operation to `newTxnRestartError`.

crdb/common.go
sravotto: My understanding is that we not calling tx.Rollback, but rather tx.Exec(ctx, "ROLLBACK"). Please check the comments in the code, and let me know if that makes sense.

crdb/retry.go
sravotto: Done.

crdb/retry.go
sravotto: Done, reworded comment.

crdb/retry_test.go
sravotto: Done.

crdb/retry_test.go
sravotto: Done.

crdb/tx.go
sravotto: Agree. Done.

crdb/retry.go
rafiss: @sravotto , how about we rename this function `Vargo` -> `ExternalBackoffPolicy`.

I think the name would be a bit more discoverable this way.

crdb/retry.go
```diff
@@ -0,0 +1,284 @@
...
+
+// VargoBackoff is an interface for external backoff strategies that provide
+// delays through a Next() method. This allows adaptation of backoff policies
+// from libraries like github.com/sethvargo/go-retry without creating a direct
+// dependency.
+//
+// Next returns the next backoff duration and a boolean indicating whether to
+// stop retrying. When stop is true, the retry loop terminates with a
+// MaxRetriesExceededError.
+type VargoBackoff interface {
```
rafiss: And along with the above, let's rename this from `VargoBackoff` -> `ExternalBackoff`.

crdb/retry.go
```diff
@@ -0,0 +1,284 @@
...
+	// Next returns the next delay duration and whether to stop retrying.
+	// When stop is true, no more retries will be attempted.
+	Next() (next time.Duration, stop bool)
+}
+
+// vargoAdapter adapts backoff policies in the style of github.com/sethvargo/go-retry.
+type vargoAdapter struct {
```
rafiss: And let's rename this `vargoAdapter` -> `externalBackoffAdapter`

crdb/retry.go
```diff
@@ -0,0 +1,284 @@
...
+	// causes early termination).
+	MaxDelay time.Duration
+}
+
+// NewRetry implements RetryPolicy.
+func (l *ExpBackoffRetryPolicy) NewRetry() RetryFunc {
+	tryCount := 0
+	return func(err error) (time.Duration, error) {
+		tryCount++
+		if tryCount > l.RetryLimit {
```
rafiss: I think we should align the behavior of `RetryLimit == 0` for `ExpBackoffRetryPolicy` with `LimitBackoffRetryPolicy`. 

What do you think about makeing `0` mean "infinite retries" for both (which is what `LimitBackoffRetryPolicy` does currently)?

crdb/retry.go
sravotto: Makes sense. Done.

crdb/retry.go
sravotto: Makes sense. Done.

crdb/retry.go
sravotto: Done.

sravotto:
Thanks for the review.

PR #184, opened 2025-06-23
Add CLI tool for downloading CockroachDB binaries with cross-platform support

rail:
- Create cmd/download-crdb-binaries CLI tool with flags:
  * -platform (linux, darwin, windows) - defaults to runtime.GOOS
  * -arch (amd64, arm64) - defaults to runtime.GOARCH  
  * -version (specific version or 'unstable') - defaults to 'unstable'
  * -output (custom directory) - defaults to temp directory

- Extend testserver package with platform-specific functions:
- Fix unstable binary filename generation:
  * Changed from 'cockroach-' to descriptive 'cockroach-latest'

- Add macOS ARM64 unsigned binary support:
  * Correct naming pattern: 'darwin-arm64.unsigned' for unstable builds
  * Enables downloading bleeding-edge builds on Apple Silicon

- Maintain backward compatibility:
  * Original functions delegate to new platform-specific versions
  * Existing test server functionality unchanged

- Remove unsupported musl detection.

The CLI tool now supports downloading any CockroachDB version for any supported platform/architecture combination, with 'unstable' as the sensible default for getting the latest bleeding-edge builds.

### DIFF
--- a/testserver/binaries.go
+++ b/testserver/binaries.go
@@ -26,10 +26,8 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 	"time"
@@ -41,8 +39,8 @@ import (
 
 const (
 	latestSuffix     = "LATEST"
-	finishedFileMode = 0555
-	writingFileMode  = 0600 // Allow reads so that another process can check if there's a flock.
+	finishedFileMode = 0o555
+	writingFileMode  = 0o600 // Allow reads so that another process can check if there's a flock.
 )
 
 const (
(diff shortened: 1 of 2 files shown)


testserver/binaries.go
```diff
@@ -137,16 +138,46 @@ func DownloadFromURL(downloadURL string) (*http.Response, error) {
...
-	dbUrl, desiredVersion, err := GetDownloadURL(desiredVersion, nonStable)
+	return DownloadBinaryWithPlatform(tc, desiredVersion, nonStable, runtime.GOOS, runtime.GOARCH, "")
+}
+
+// DownloadBinaryWithPlatform saves the specified version of CRDB for the given
+// platform and architecture into a local binary file, and returns the path for
+// this local binary.
+// To download a specific cockroach version, specify desiredVersion. Otherwise,
+// the latest stable or non-stable version will be chosen.
+// To download the latest STABLE version of CRDB, set `nonStable` to false.
```
rickystewart, line 149: Nit: having the variable be `nonStable` instead of `stable` is a bit unintuitive. I think the fact that you need to add this clarification to the function comment here is evidence for this -- you need to be very careful to explain what the parameter does, when if it were just `stable bool`, then it would be obvious.

rickystewart (approved):

testserver/binaries.go
rail: I just cargo-culted the variable name. Let me file a ticked to rename it.

PR #186, opened 2025-06-26
testserver: add option to run demo mode

benbardin:
The hard-coded defaults here around job schedules are more reasonable for some unit tests.


testserver/binaries.go
```diff
@@ -403,7 +403,6 @@ func downloadBinaryFromTar(response *http.Response, output *os.File, filePath st
 		}
 
 	}
-	return nil
```
benbardin, line 406: unreachable code.

testserver/testserver.go
```diff
@@ -559,12 +566,7 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
 		return nil, fmt.Errorf("%s failed to parse version: %w", testserverMessagePrefix, err)
 	}
 
-	startCmd := "start-single-node"
```
benbardin, line 562: moved into the new `else` below

testserver/testserver.go
```diff
@@ -573,60 +575,100 @@ func NewTestServer(opts ...TestServerOpt) (TestServer, error) {
...
-				"--listening-url-file=" + nodes[i].listeningURLFile,
-				"--external-io-dir=" + serverArgs.externalIODir,
+			if serverArgs.storeOnDisk {
+				storeArg = fmt.Sprintf("--store=path=%s", nodeBaseDir)
+			}
+
+			// TODO(janexing): Make sure the log is written to logDir instead of shown in console.
+			// Should be done once issue #109 is solved:
+			// https://github.com/cockroachdb/cockroach-go/issues/109
+			nodes[i].stdout = filepath.Join(logsBaseDir, "cockroach.stdout")
```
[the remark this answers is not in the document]
benbardin: Demo mode and regular mode have different flags, so I think a helper here might be more risky than, well, helpful. I'll add a comment in the code.

benbardin:
I'll add your suggestion to the unreachable-code thing, that makes sense.

PR #187, opened 2025-09-04
demo: fix socket location for bazel sandbox

benbardin:
This resolves a failure when demo mode tried to start a socket in $HOME without permissions.

### DIFF
--- a/testserver/testservernode.go
+++ b/testserver/testservernode.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"syscall"
 )
@@ -103,7 +104,6 @@ func (ts *testServerImpl) StartNode(i int) error {
 		"COCKROACH_MAX_OFFSET=1ns",
 		"COCKROACH_TRUST_CLIENT_PROVIDED_SQL_REMOTE_ADDR=true",
 	}
-	currCmd.Env = append(currCmd.Env, ts.serverArgs.envVars...)
 
 	// Set the working directory of the cockroach process to our temp folder.
 	// This stops cockroach from polluting the project directory with _dump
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-go/187)* status: <span>:shipit:</span> complete! 0 of 0 LGTMs obtained (waiting on @rickystewart)

___
*[`testserver/testservernode.go` line 145 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/187#-OZKSbFT-ITvA153RP_i:-OZKSbFT-ITvA153RP_j:b-wy9gf9) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/fcc31ab0146f6ebda61471b76cea2357a4c24bac/testserver/testservernode.go#L145)):*
> ```Go
> 	if ts.serverArgs.demoMode {
> 		// Prefer a very short path to stay under Unix-domain socket length limits.
> 		preferred := filepath.Join("/tmp", fmt.Sprintf("cd-%d", os.Getpid()))
> ```

[nit] We should use [os.MkdirTemp](https://pkg.go.dev/os#MkdirTemp)
___
*[`testserver/testservernode.go` line 148 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-go/187#-OZKSwKSEMyNNs9qnm_L:-OZKSwKSEMyNNs9qnm_M:b-m641lh) ([raw file](https://github.com/cockroachdb/cockroach-go/blob/fcc31ab0146f6ebda61471b76cea2357a4c24bac/testserver/testservernode.go#L148)):*
> ```Go
> 		demoHome := preferred
> 		if err := os.MkdirAll(demoHome, 0700); err != nil {
> 			log.Printf("%s: /tmp not writable for demo HOME (%v); falling back to base dir %s", testserverMessagePrefix, err, ts.baseDir)
> ```

Is this necessary (did you see this error under bazel)?

rickystewart (approved):

